PR #28, opened 2026-01-29
fix(agentmail): handle timestamps without timezone suffix

rawwerks:
## Summary
- Adds `FlexTime` type to handle both RFC3339 and timezone-less ISO8601 timestamps from Agent Mail server
- Handles wrapped/raw array formats in `ListProjectAgents`

## Test plan
- [x] Verified parsing of timestamps with timezone suffix (RFC3339)
- [x] Verified parsing of timestamps without timezone suffix (ISO8601)
- [x] Verified `ListProjectAgents` handles both wrapped and raw array formats

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/internal/agentmail/tools.go
+++ b/internal/agentmail/tools.go
@@ -713,8 +713,21 @@ func (c *Client) ListProjectAgents(ctx context.Context, projectKey string) ([]Ag
 		return []Agent{}, nil
 	}
 
+	// The server may return either:
+	// 1. A wrapped response: {"project": {...}, "agents": [...]}
+	// 2. A raw array: [...]
+	// Try wrapped format first, fall back to raw array.
+	var wrappedResp struct {
+		Agents []Agent `json:"agents"`
+	}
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks for the contribution! Per project policy, I don't merge outside PRs, but I've reviewed your approach and it addresses real issues.

The problems you identified are valid:
1. **Timestamp parsing** — The Agent Mail server can return ISO8601 timestamps without timezone suffixes, which breaks `time.Parse(time.RFC3339, ...)`. Your `FlexTime` approach of trying multiple layouts is reasonable.
2. **ListProjectAgents response format** — The server may return wrapped `{"agents": [...]}` or raw `[...]` arrays, and only handling one breaks the other.

I'll implement fixes for both independently. The timestamp handling can likely be simplified with a single fallback parse rather than iterating through multiple layouts, and the response format handling is straightforward.

Appreciate you tracking down these edge cases — they're the kind of interop issues that only surface in real usage.

PR #29, opened 2026-01-30
fix(activity): idle prompts take priority over historical errors

rawwerks:
## Summary

- Fix false ERROR state in `ntm activity` for Codex (and other) agents
- Idle prompts now take priority over error patterns when velocity is low
- Adds regression test for the specific scenario

## Problem

When running `ntm activity --cod`, Codex agents were incorrectly shown as ERROR state when they were actually healthy and waiting at their prompt. This happened because:

1. `classifyState()` checked error patterns **before** idle patterns
2. Error patterns matched historical output (e.g., "failed to find file" from earlier)
3. Even with an idle prompt visible, ERROR took priority

## Solution

Reordered the priority in `classifyState()`:
1. Check for idle prompt + low velocity FIRST
2. If agent shows idle prompt and is not actively outputting, return WAITING
3. Only check error patterns when NOT at an idle prompt

This matches user expectations: if an agent is at a prompt ready for input, it's WAITING regardless of what error messages appear earlier in the scrollback.

## Test plan

- [x] Added unit test `TestClassifyState_IdleTakesPriorityOverHistoricalErrors`
- [x] All existing tests pass
- [ ] Manual test with live Codex agent

### DIFF
--- a/internal/robot/activity.go
+++ b/internal/robot/activity.go
@@ -509,14 +509,9 @@ func (sc *StateClassifier) Classify() (*AgentActivity, error) {
 // classifyState determines state based on velocity and patterns.
 // Returns state, confidence, and trigger description.
 func (sc *StateClassifier) classifyState(velocity float64, matches []PatternMatch) (AgentState, float64, string) {
-	// Error patterns take priority
-	for _, m := range matches {
-		if m.Category == CategoryError {
-			return StateError, 0.95, "error_pattern:" + m.Pattern
-		}
-	}
-
-	// Check for idle prompt with low velocity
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks for the contribution! Per project policy, I don't merge outside PRs, but I reviewed your approach.

The core insight is correct: when an agent is showing an idle prompt and has low velocity, it should be classified as WAITING regardless of historical error messages in the scrollback. The current code checks errors first, which can cause false positives when old errors remain visible.

I'll implement a similar fix independently. The PR also included unrelated changes (FlexTime refactor, FormatPaneTarget changes) which I'll evaluate separately.

Appreciate you taking the time to identify and document the issue!

PR #41, opened 2026-02-03
fix: scrollback capture pane ID + stabilize send test

### LINKED ISSUE #39: bug: scrollback capture uses pane.Index instead of pane.ID
## Summary

`captureScrollbackEnhanced` in `internal/checkpoint/scrollback.go` passes `pane.Index` (an integer like `2`) instead of `pane.ID` (a tmux pane ID like `%123`) to `CaptureScrollback`. This causes tmux to interpret the target as a window number rather than a pane ID, resulting in "can't find window" errors.

## Root Cause

**Introduced in:** `11c2ea1` (2026-01-06)
**Commit message:** `feat(checkpoint): add gzip compression for scrollback capture (ntm-1l9t)`

The bug exists from the initial implementation:

```go
// scrollback.go:191
capture, err := CaptureScrollback(cp.SessionName, fmt.Sprintf("%d", pane.Index), config)

// scrollback.go:75
target := fmt.Sprintf("%s:%s", session, paneID)  // Results in "session:2"
```

The `Pane` struct has two relevant fields:
- `ID string` - tmux pane ID (e.g., `%123`) - globally unique
- `Index int` - pane index within window (e.g., `2`)

When `pane.Index` is used, tmux interprets `session:2` as "window 2 in session", not "pane with index 2".

## Error Message

```
WARN failed to capture scrollback pane=2 error="capturing pane output: 
/usr/bin/tmux capture-pane -t session:2 -p -S -500: exit status 1: can't find window: 2"
```

## Fix

Use `pane.ID` directly (tmux pane IDs are globally unique and don't need session prefix):

```go
capture, err := CaptureScrollback(cp.SessionName, pane.ID, config)
```

## Why Not Caught

No integration test exists for `captureScrollbackEnhanced` with a real tmux session. Existing tests only cover compression/decompression and storage operations.

## Affected Functionality

- Checkpoint scrollback capture
- Any feature relying on `checkpoint.Capturer.captureScrollbackEnhanced`

### LINKED ISSUE #40: test: TestSendRealSession flaky due to cat alias and timing
## Summary

`TestSendRealSession` in `internal/cli/send_test.go` is flaky on systems where `cat` is aliased to `bat` and when shell initialization takes longer than 500ms.

## Root Cause

**Introduced in:** `3f8c7e2` (2025-12-10)
**Commit message:** `test(cli): add real-tmux integration test for send command`

### Problem 1: `cat` alias

```go
cfg.Agents.Claude = "cat"  // Can be aliased to "bat"
```

On many systems (especially with modern shell configs), `cat` is aliased to `bat`:
```bash
$ type cat
cat is an alias for bat
```

`bat` has different stdin handling behavior than `/bin/cat`:
- Shows interactive STDIN preview
- May buffer input differently
- Has initialization overhead

### Problem 2: Insufficient wait time

```go
// Wait for session to settle
time.Sleep(500 * time.Millisecond)
```

500ms is not always sufficient for:
1. Shell initialization (loading `.zshrc`, plugins)
2. Processing `cd /path/to/project` command
3. zsh auto-ls feature (if enabled) - outputs directory listing
4. `bat` initialization

### Observed Behavior

Test sends prompt before the agent pane is ready to receive input:

```
Pane output did not contain prompt "Hello NTM Test". Got:
    cd '/tmp/...' && NTM_SPAWN_ORDER=1 ... cat
    ─────┬─────────────────────────────
         │ STDIN
    ─────┼─────────────────────────────
       1 │
       2 │
```

The bat STDIN display shows empty lines because the prompt was sent before bat was ready.

## Fix

1. Use `/bin/cat` explicitly to avoid alias:
```go
cfg.Agents.Claude = "/bin/cat"
```

2. Increase settle time:
```go
time.Sleep(1500 * time.Millisecond)
```

## Reproduction

Only reproduces on systems where:
- `cat` is aliased to `bat`
- Shell has auto-ls on cd or heavy `.zshrc`

The test passes when run as part of the full suite (due to timing/caching) but fails when run individually with `-run TestSendRealSession`.

joyshmitz:
## Summary

Two bug fixes:

1. **Scrollback capture used wrong pane identifier** - `captureScrollbackEnhanced` passed `pane.Index` (integer) instead of `pane.ID` (tmux ID like `%123`), causing "can't find window" errors
2. **Flaky TestSendRealSession** - `cat` alias to `bat` + insufficient settle time caused intermittent failures

### `internal/checkpoint/scrollback.go`
- Use `pane.ID` instead of `fmt.Sprintf("%d", pane.Index)`
- Remove unnecessary `session:` prefix (tmux pane IDs are globally unique)

### `internal/cli/send_test.go`
- Use `/bin/cat` explicitly to avoid shell aliases
- Increase settle time from 500ms to 1500ms

## Test Results

```
=== Run 1 === ok  github.com/Dicklesworthstone/ntm/internal/cli  6.836s
=== Run 2 === ok  github.com/Dicklesworthstone/ntm/internal/cli  6.840s
=== Run 3 === ok  github.com/Dicklesworthstone/ntm/internal/cli  6.829s
```

Fixes #39, Fixes #40

### DIFF
--- a/internal/cli/send_test.go
+++ b/internal/cli/send_test.go
@@ -41,8 +41,9 @@ func TestSendRealSession(t *testing.T) {
 
 	// Use a simple echo command that persists for a bit so we can capture it
 	// We use 'read' to keep the pane open/active if needed, or just sleep
-	cfg.Agents.Claude = "cat" // Simple cat will echo whatever we send to stdin/tty?
-	// Actually, SendKeys sends keystrokes. "cat" will print them back. Perfect.
+	// Use /bin/cat explicitly to avoid shell aliases (e.g., cat -> bat) which
+	// have different input/output behavior and can cause test flakiness.
+	cfg.Agents.Claude = "/bin/cat"
 
 	sessionName := fmt.Sprintf("ntm-test-send-%d", time.Now().UnixNano())
 	defer func() {
@@ -73,8 +74,11 @@ func TestSendRealSession(t *testing.T) {
 		t.Fatalf("spawnSessionLogic failed: %v", err)
 	}
 
-	// Wait for session to settle
-	time.Sleep(500 * time.Millisecond)
+	// Wait for session to settle - needs enough time for:
+	// 1. Shell to initialize (load zshrc, etc.)
+	// 2. The cd && cat command to be processed
+	// 3. cat to be ready to receive input
+	time.Sleep(1500 * time.Millisecond)
 
 	// Send a prompt
 	prompt := "Hello NTM Test"


Dicklesworthstone:
Thanks for the contribution! Per project policy, I don't merge outside PRs, but I reviewed your approach.

The fixes look solid:
1. Using `pane.ID` instead of `pane.Index` for scrollback capture - correct, pane IDs are globally unique
2. Using `/bin/cat` explicitly to avoid shell aliases - good defensive coding
3. Increasing settle time from 500ms to 1500ms - reasonable for shell initialization

I'll implement similar fixes independently. Appreciate the thorough bug reports in #39 and #40 as well!

PR #120, opened 2026-05-04
Add force non-interactive send flag

### LINKED ISSUE #119: ntm send: add --force-non-interactive for recovery/status automation
## Problem

Automation that performs protected-session recovery currently has to use shell-level confirmation pipes around `ntm send` when a duplicate-work/CASS confirmation can appear. This is fragile for recovery tooling because a stuck interactive confirmation can prevent a worker relaunch/status prompt from landing.

## Current code path observed

- `internal/cli/send.go` exposes `--no-cass-check`, but not a broader non-interactive send mode.
- When similar CASS work is found, `runCassDuplicateCheck` prompts `Continue anyway?` in interactive mode.
- `confirm_huh.go` falls back to reading `y`/`yes` from stdin when not in a TTY.
- `getPromptContent` gives positional arguments and `--file` priority over stdin, which lets wrappers pipe `y` while still passing a prompt.

## Requested flag

Add a narrow force flag to `ntm send`:

```bash
ntm send <session> --pane=<pane> --force-non-interactive "<prompt>"
```

Suggested behavior:

- Never block on an interactive confirmation.
- Bypass confirmation gates that are safe for automated recovery/status sends, starting with CASS duplicate confirmation.
- Fail closed on destructive or ambiguous confirmation classes.
- Emit a JSON field such as `non_interactive_forced=true` when used.
- Keep `--no-cass-check` as the explicit CASS-only bypass; `--force-non-interactive` should be the mechanical wrapper-friendly contract.

## Current local workaround

Protected-session recovery currently uses:

```bash
printf 'y\n' | ntm send "$session" --pane="$pane" --no-cass-check "$prompt" \
  || yes | ntm send "$session" --pane="$pane" --no-cass-check "$prompt"
```

This works because prompt content is positional and stdin remains available for confirmation reads, but it is not a durable CLI contract.

zerone0x:
## Summary

- Add `ntm send --force-non-interactive` as a mechanical wrapper-friendly flag.
- Use it to bypass the CASS duplicate `Continue anyway?` confirmation while leaving `--no-cass-check` intact as the explicit CASS-only bypass.
- Emit `non_interactive_forced: true` in send JSON outputs when the flag is used.
- Add tests for flag parsing and JSON field emission.

## Scope / safety

Orbit assessment: medium/hunter-pro due CLI/CASS send-path tracing, but the implementation is intentionally narrow and safe: it only bypasses the CASS duplicate confirmation. Other confirmation classes are unchanged.

Fixes #119.

## Verification

- `git diff --check` passed.
- Targeted `go test -run 'TestSendForceNonInteractiveFlagParses|TestSendJSONIncludesNonInteractiveForcedOnlyWhenTrue' ./internal/cli` could not run in this environment because `go`/`gofmt` are not installed (`go: command not found`).

### DIFF
--- a/internal/cli/send.go
+++ b/internal/cli/send.go
@@ -50,6 +50,7 @@ type SendResult struct {
 	Success       bool               `json:"success"`
 	Session       string             `json:"session"`
 	PromptPreview string             `json:"prompt_preview,omitempty"`
+	NonInteractiveForced bool        `json:"non_interactive_forced,omitempty"`
 	Redaction     *RedactionSummary  `json:"redaction,omitempty"`
 	Warnings      []string           `json:"warnings,omitempty"`
 	Blocked       bool               `json:"blocked,omitempty"`
@@ -76,6 +77,7 @@ type SendDryRunResult struct {
 	Success   bool               `json:"success"`
 	DryRun    bool               `json:"dry_run"`
 	Session   string             `json:"session"`
+	NonInteractiveForced bool    `json:"non_interactive_forced,omitempty"`
 	Redaction *RedactionSummary  `json:"redaction,omitempty"`
 	Warnings  []string           `json:"warnings,omitempty"`
 	Blocked   bool               `json:"blocked,omitempty"`
@@ -233,9 +235,10 @@ type SendOptions struct {
 	RouteStrategy string // Routing strategy (least-loaded, round-robin, etc.)
 
 	// CASS check options
-	CassCheck      bool
-	CassSimilarity float64
-	CassCheckDays  int
+	CassCheck             bool
+	CassSimilarity        float64
+	CassCheckDays         int
+	ForceNonInteractive   bool
 
 	// Hooks
 	NoHooks bool
@@ -488,6 +491,7 @@ func newSendCmd() *cobra.Command {
 	var dryRun bool
 	var cassCheck bool
 	var noCassCheck bool
+	var forceNonInteractive bool
 	var cassSimilarity float64
 	var cassCheckDays int
 	var noHooks bool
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks for the patch and the targeted test design — your shape was exactly right (single `force` flag threaded through to `checkCassDuplicates`, JSON field with `omitempty` so default callers don't see a contract change). Per project policy I don't merge outside PRs directly, but I implemented the same fix independently in [`1ab2e2be3e9a1fa1ae18b4e767add7fb68285eca`](https://github.com/Dicklesworthstone/ntm/commit/1ab2e2be3e9a1fa1ae18b4e767add7fb68285eca). Same surface (`--force-non-interactive`, `non_interactive_forced` JSON field) with a few hardening tweaks:

- The bypass-warning is written to **stderr** rather than stdout so it doesn't pollute machine-readable JSON output (your version put it on stdout via `fmt.Printf`, which would interleave into a JSON pipeline).
- Help text explicitly documents which confirmation classes are bypassed (CASS only, today) and which fail closed (destructive/ambiguous), so the contract is auditable rather than something callers have to derive from reading source.
- Test coverage is broader: registration + default + parsing + help-text contract + JSON-shape assertions across `SendResult`, `SendDryRunResult`, and `BatchResult` (your test set covered the first and last; I added the `omitempty` regression guard so default callers can't silently see the field appear).

Verified: `go build ./... && go test ./internal/cli/ -run 'TestSend' -count=1` clean. Closes #119.

Appreciate the careful framing in your PR description (orbit assessment, scope/safety note) — that made this a one-pass implementation rather than triage thrash.

PR #156, opened 2026-05-21
status: scan 20 lines instead of 3 when detecting idle prompts

IvanWest33:
## Summary

`DetectIdleFromOutput` in `internal/status/patterns.go` was capped at scanning the last 3 non-empty lines of the captured pane buffer when looking for an idle-prompt pattern. In narrow tiled-pane swarm setups (16+ tiled panes per session), the 3-line window is too small — the bottom 3 lines are dominated by Claude Code's persistent status bar / hint / separator / input-box border, and the actual `❯ ` prompt sits ~5–15 lines above. The pane is then misclassified as `ActivityActive` indefinitely.

This PR extracts the limit to a named constant `maxIdleDetectionLines = 20`, sized to cover Claude Code's typical footer with headroom while staying well short of the 50-line health-capture buffer (`internal/health/health.go:233`).

## Reproduction (without this fix)

1. `ntm spawn myproject --cc=16 --layout=tiled` (or similar that yields per-pane size ≤ 18 chars × ~4–6 lines)
2. Wait for every Claude Code pane to finish session-restore — every pane is now at the `❯` empty prompt
3. `ntm health myproject --json`

**Expected:** most/all panes report `activity: "idle"`
**Actual (before fix):** 15 of 16 panes report `activity: "active"`, `idle_seconds: 0`. The bottom 3 lines captured by `health.go:233 CapturePaneOutputContext(..., 50)` are CC's status bar:
```
───────────
  ⏵⏵ bypass permissions on (shift+tab to cycle)
  new task? /cl…
```
None of these match `IsPromptLine`, so `DetectIdleFromOutput` returns false → `detectActivity` falls through to `ActivityActive` since `idleTime < 5*time.Minute`.

## Why this matters

`activity == "idle"` is the canonical signal for downstream dispatchers (including external orchestrators on top of NTM) to identify panes safe to send work to. When ~31 of 32 panes are wrongly reported active, dispatch logic refuses to fan out and the swarm parks at single-digit concurrent throughput despite all panes being unambiguously idle.

Real-world impact on our setup: a 32-pane swarm (`--cc=16 --cod=16`) was running at ~5 concurrent beads in_progress for hours after every dispatch, until we patched our own controller to bypass the `activity` field. Tracing the root cause through NTM source led to the 3-line constant.

## Fix

- Replace `const maxLinesToCheck = 3` (local to `DetectIdleFromOutput`) with a package-level `maxIdleDetectionLines = 20`.
- Update the function doc comment with the rationale (narrow tiled panes, CC footer geometry).
- Existing tests `prompt in third line from end` and "prompt 5 lines from end" (former "prompt beyond 3 lines", originally `expected: false`) continue to pass with the larger window.
- New regression test `narrow tiled cc pane with footer noise` captures the exact failure mode.
- New `prompt beyond 20 lines` test guards against false positives from very-old prompt text in scrollback.

## Tests

```
$ go test ./internal/status/...
ok  	github.com/Dicklesworthstone/ntm/internal/status	56.856s
$ go test ./internal/health/...
ok  	github.com/Dicklesworthstone/ntm/internal/health	0.038s
```

All package tests pass — the larger window does not flip any previously-passing test.

## Alternatives considered

- **Dynamic window from `pane.Height`** — more principled but requires plumbing pane geometry through the capture path; deferred as a follow-up if maintainers prefer.
- **Smarter regex matching that ignores known status-bar lines** — fragile, pattern-specific, requires per-agent maintenance.
- **Increase to a much higher number (e.g. 50)** — 20 was chosen as the smallest value that comfortably covers the observed CC footer (6–12 lines) without scanning the entire capture buffer.

Happy to iterate on the constant value or approach if maintainers have a preference. Setup: NTM v1.18.1 baseline, Go 1.26.2, tmux 3.x, Claude Code 2.x, Ubuntu 24.04.

### DIFF
--- a/internal/status/patterns.go
+++ b/internal/status/patterns.go
@@ -151,8 +151,18 @@ var knownAgentTypes = map[string]bool{
 var knownAgentPromptPrefixes = regexp.MustCompile(`(?i)^(claude|codex|gemini|cursor|windsurf|aider|ollama)>\s*$`)
 
 // DetectIdleFromOutput analyzes output to determine if agent is idle.
-// It checks up to 3 non-empty lines from the end for prompt patterns.
-// This window allows detecting idle state even when there's trailing output.
+// It checks up to maxIdleDetectionLines non-empty lines from the end for prompt
+// patterns. This window allows detecting idle state even when the agent's TUI
+// renders multiple lines of status / hint / separator below the actual prompt
+// line — which is common in narrow tiled panes where Claude Code's status bar,
+// thinking-budget indicator, and input-box border together push the `❯` prompt
+// 5–15 visible lines above the last line of output.
+//
+// The previous 3-line window was too narrow for swarm setups with many tiled
+// panes (e.g. 16+ panes per session yielding ~18-char-wide × ~4–6-line-tall
+// viewports). In that geometry the bottom 3 non-empty lines are dominated by
+// the persistent status bar and the actual idle prompt is never inspected,
+// causing the pane to be misclassified as ActivityActive forever.
 func DetectIdleFromOutput(output string, agentType string) bool {
 	agentType = string(agent.AgentType(agentType).Canonical())
 
@@ -164,11 +174,8 @@ func DetectIdleFromOutput(output string, agentType string) bool {
 		return false
 	}
 
-	// Check up to 3 non-empty lines from the end for prompt patterns.
-	// This window allows detecting idle state even with some trailing output.
-	const maxLinesToCheck = 3
 	linesChecked := 0
-	for i := len(lines) - 1; i >= 0 && linesChecked < maxLinesToCheck; i-- {
+	for i := len(lines) - 1; i >= 0 && linesChecked < maxIdleDetectionLines; i-- {
 		line := strings.TrimSpace(lines[i])
 		if line == "" {
 			continue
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thank you for this — the diagnosis is excellent, and the reproduction (narrow tiled panes where Claude Code's persistent footer pushes the `❯` prompt past the bottom 3 lines) pinpointed a real bug in `DetectIdleFromOutput`. The trace through to the `maxLinesToCheck = 3` constant and the downstream impact on the canonical `activity == "idle"` signal was spot-on.

A note on our process: as a matter of policy we do not merge outside PRs — we treat them as intel and reimplement independently so the maintainers retain full authorship and responsibility for every line. So rather than merge this, we used it to drive our own fix, which has now landed in `e28763ea` (released as `v1.18.2`).

On the approach itself: we went a step beyond a plain widen to 20 lines, because a wider window alone is unsafe. Claude Code keeps its input box (`❯ `) rendered **while it is actively working** — the spinner draws above it, the box stays. A bare scan would therefore match that persistent prompt mid-task and flip a busy pane to `idle`. This is the exact false-idle class the maintainers previously fixed by removing the `"bypass permissions on"` status-bar pattern, and that the existing `internal/agent.detectIdle` already guards against with a "most-recent-on-screen-marker-wins" ordering rule.

So our implementation:
- Widens the scan window (to 12 lines, matching the empirically-validated CC footer window already used in `internal/agent.detectIdle`, and well short of the 50-line health capture buffer so stale scrollback prompts are not matched).
- Adds a **spinner-after-prompt guard**: when a prompt is matched, the idle verdict is rejected if an active-work spinner (`Scurrying… (12s)`, `· thinking`, `Running…`, `esc to interrupt`) appears on a line *below* the prompt. This keeps actively-working panes correctly `active` while still surfacing genuinely-idle panes that were previously stuck.
- Adds regression tests for the narrow-tiled-pane idle case, the working-with-spinner-below case (must stay active), a stale-spinner-above-prompt case (idle), and a beyond-window guard.

Closing as implemented independently in `e28763ea`. Genuinely appreciate the careful write-up and the alternatives section — it made the right design obvious.

PR #160, opened 2026-05-24
fix(assign): treat --agent any as no provider filter

IvanWest33:
## Summary

- Treat `--agent any`, `--agent all`, and `--agent *` as no provider filter for `ntm assign`.
- Reuse the same normalization in both the enhanced assign scan and `getIdleAgents`.
- Add a regression test so `any` cannot silently become a literal provider name again.

## Root cause

`--agent any` was passed through `ResolveAgentType`, which returned the literal string `any`. The enhanced assign path then compared each pane provider (`claude`, `codex`, etc.) against `any`, so every pane was filtered out before idle detection ran. In mixed-provider sessions this made `idle_agent_count` report `0` even while health/activity surfaces showed idle panes.

## Impact

Controller/watch flows configured with `--agent any` can now dispatch to all eligible providers as intended. Existing provider-specific filters (`claude`, `codex`, aliases like `cod`) keep their previous behavior.

## Validation

- `go test ./internal/cli ./internal/robot ./internal/agent`
- Live sanity check in a mixed Claude/Codex session: `ntm assign ... --agent any --dry-run --json` went from `idle_agent_count: 0` to matching the all-provider idle pool.

Kept intentionally small and easy to cherry-pick/squash/reword if you prefer to land it manually.

### DIFF
--- a/internal/cli/assign.go
+++ b/internal/cli/assign.go
@@ -201,7 +201,7 @@ Examples:
 	cmd.Flags().IntVar(&assignLimit, "limit", 0, "Maximum number of assignments (0 = unlimited)")
 
 	// Agent type filters
-	cmd.Flags().StringVar(&assignAgentType, "agent", "", "Filter by agent type: claude, codex, gemini")
+	cmd.Flags().StringVar(&assignAgentType, "agent", "", "Filter by agent type: any, claude, codex, gemini")
 	cmd.Flags().BoolVar(&assignCCOnly, "cc-only", false, "Only assign to Claude agents (alias for --agent=claude)")
 	cmd.Flags().BoolVar(&assignCodOnly, "cod-only", false, "Only assign to Codex agents (alias for --agent=codex)")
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks for the PR! Per project policy I don't merge outside contributions — I do read every diff, though, and this is a good catch.

The root cause analysis is correct: `robot.ResolveAgentType("any")` returns the literal `"any"`, and the enhanced assign path then filters out every pane provider that doesn't equal that literal. The aliases `--agent all` / `--agent *` are a nice ergonomic addition I hadn't considered.

I'll implement this independently — same shape (normalize `any|all|*|""` to "no filter" in a shared helper, reuse from both `resolveAgentTypeFilter` and `getIdleAgents`, add a regression test that covers the matrix). Closing this PR per policy; tracking the work separately so you'll see the fix land on `main` shortly.

Appreciate the time you took.

PR #161, opened 2026-05-24
[codex] filter non-ready assign recommendations

IvanWest33:
## Summary
- filter `bv --robot-triage` recommendations before `ntm assign` turns them into dispatchable beads
- skip dependency-blocked beads, non-open statuses, operator-gated labels, and beads already active in the assignment store
- keep `actionable_count`/`blocked_count` aligned with assignable work instead of counting stale recommendations as ready

## Root Cause
`getAssignOutputEnhanced` trusted `bv --robot-triage` recommendations and only checked `blocked_by`. If triage output contained `blocked`, `in_progress`, operator-gated, or already-assigned beads, `ntm assign --dry-run` could report them as actionable and could dispatch them in auto mode.

## Validation
- `go test ./internal/cli -run 'TestFilterAssignableTriageRecommendationsRejectsNonReadyItems|TestFilterAssignedReadyBeadsRejectsActiveStoreItems|TestDependencyFilteringInAssignment|TestAssignAgentTypeFilter'`
- `go test -timeout 20s ./internal/cli -run 'TestFilterAssignableTriageRecommendationsRejectsNonReadyItems|TestFilterAssignedReadyBeadsRejectsActiveStoreItems'`
- `go build -o /tmp/ntm-assign-filter-fix ./cmd/ntm`


Dicklesworthstone:
Thanks for the PR! Per project policy I don't merge outside contributions, but the diff is informative and I appreciate the careful filter taxonomy.

Useful intel from this change:

1. `bv --robot-triage` recommendations are not safe to trust as "actionable" — they include `blocked` / `in_progress` / closed / operator-gated beads, and the current `getAssignOutputEnhanced` only filters on `blocked_by`. That gap is real.
2. Cross-checking against the local assignment store to filter `already_assigned` is the right defense — `bv` doesn't know what's been claimed in this session.
3. The structured `SkippedItem.Reason` taxonomy (`blocked_by_dependency`, `blocked_status`, `already_in_progress`, `operator_gated`, `already_assigned`, `not_open_status`) is a nicer surface than the current binary blocked/ready split, and the `countSkippedByReason` helper keeps `BlockedCount` semantically correct after the broadening.

I'll implement these independently with my own approach (likely: stricter `bv` contract upstream so we don't double-filter, plus a smaller assignment-store join that doesn't drift if assignment.LoadStore signature changes). Closing this PR per policy.

Operator-gated label list and the test-table shape will probably come over close to verbatim — those are well-chosen — though I'll have to attribute the policy taxonomy to my own design decisions rather than your enumeration. Appreciate the work.

PR #164, opened 2026-05-30
bd-exoe6: skip blocked beads in assign recommendations

IvanWest33:
This PR addresses the controller/assigner regression where blocked parent beads could be proposed as assignable during dependency-driven dry-runs.

Changes:
- exclude recommendations with status=blocked even when blocked_by is empty
- keep dependency-blocked recommendations skipped with an explicit blocked reason
- add regression coverage for blocked-status and blocked-by filtering

Verified locally:
- go test ./internal/bv -run TestBVClientConvertRecommendation -count=1
- go test ./internal/cli -run TestPartitionAssignableRecommendationsSkipsBlockedStatusAndDependencies -count=1
- ntm assign notaryware --dry-run --strategy=dependency --agent codex --limit=4 --json

Upstream push from this shell is read-only on Dicklesworthstone/ntm, so the code is published on fork branch IvanWest33/ntm:bd-exoe6.

### DIFF
--- a/internal/bv/client.go
+++ b/internal/bv/client.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"os/exec"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -349,7 +350,7 @@ func (c *BVClient) convertRecommendation(rec TriageRecommendation) Recommendatio
 		UnblocksCount: len(rec.UnblocksIDs),
 		UnblocksIDs:   rec.UnblocksIDs,
 		BlockedByIDs:  rec.BlockedBy,
-		IsActionable:  len(rec.BlockedBy) == 0,
+		IsActionable:  IsActionableRecommendation(rec),
 		Tags:          rec.Labels,
 		Score:         rec.Score,
 		Action:        rec.Action,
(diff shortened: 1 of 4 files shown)


Dicklesworthstone:
Thanks for this, @IvanWest33 — and for the clear repro + tests. The diagnosis is correct. Per CONTRIBUTING.md I don't merge outside PRs directly, but I verified it against `main` and independently implemented the live half:

- The `internal/cli/assign.go` change here is already superseded: `2efe2ac4` replaced that open-coded blocked-filter loop with `classifyTriageRecForAssignment`, which already skips `status=blocked` (plus `in_progress`/`closed`/`operator_gated`/`already_assigned`). So that portion would conflict/regress.
- But you were right about the second half: `convertRecommendation` in `internal/bv/client.go` still set `IsActionable: len(rec.BlockedBy) == 0`, ignoring `status=blocked`. That's a real, still-live gap that surfaces over the `/api/v1/beads/recommend` serve endpoint and the `FilterReady` path, independent of the assign fix.

Fixed independently in 9353cd50: added a status-aware `isActionableRecommendation` helper (the `cli` classifier can't be imported into the `bv` package without a cycle) plus a table-driven regression test. Closing this PR — it directly drove the fix. Appreciated!

PR #174, opened 2026-06-06
Fix scanner stdout pipe race in hook scan

profplum700:
## Summary

Fixes a scanner pipe lifecycle failure in `internal/scanner/scanner.go` where NTM could abort pre-commit scans with `reading output: read |0: file already closed` when staged source files triggered UBS scanning.

The scanner no longer uses `cmd.StdoutPipe` plus a manual read path. Instead it attaches a bounded stdout writer to the command and uses `cmd.Run()`, so stdout is drained while the process runs and `Run` waits only after the copy completes. The existing `MaxScanOutputBytes` cap is preserved by a small writer that records overflow without returning a write error to `os/exec`'s copy goroutine.

This also aligns the pre-commit hook's CLI/default timeout to 120s, matching NTM's scanner config default. On a downstream Next.js project reproduction in the local environment, the staged source-file scan completed cleanly in about 70-80 seconds, so the prior 60s hook default was too short for that real guard path.

## Verification

- `go test ./internal/scanner -run 'Test(IsAvailable|DefaultOptions|BuildArgs|Parse|ScanResult|Format|Severity|Compute|Analyze|Truncate|Sort|Shorten|Config|Extract)'`
- `go test ./internal/hooks -run PreCommit`
- `make build`
- `./ntm version`
- Real downstream reproduction in a Next.js project: staged `.tsx` source file no longer fails pre-commit with `file already closed` and exits 0 with the default hook timeout.
- Normal guarded git commit without `--no-verify` succeeds for a staged `.tsx` file.
- Beads-only guarded commit succeeds.

Note: local `go test ./internal/scanner` exercises real UBS scans and timed out in the local environment (`TestAutoScanner_TriggerScan`, `TestScanDirectory`), so targeted pure scanner tests plus the downstream hook reproduction were used as acceptance proof.

Posted by Codex on behalf of profplum700.

### DIFF
--- a/internal/cli/hooks.go
+++ b/internal/cli/hooks.go
@@ -294,7 +294,7 @@ This is also called by the installed hook script.`,
 
 	cmd.Flags().BoolVarP(&verbose, "verbose", "v", false, "Verbose output")
 	cmd.Flags().BoolVar(&failOnWarning, "fail-on-warning", true, "Fail on warnings")
-	cmd.Flags().IntVar(&timeout, "timeout", 60, "Timeout in seconds")
+	cmd.Flags().IntVar(&timeout, "timeout", 120, "Timeout in seconds")
 
 	return cmd
 }
--- a/internal/hooks/hooks_test.go
+++ b/internal/hooks/hooks_test.go
@@ -151,8 +151,8 @@ func TestDefaultPreCommitConfig(t *testing.T) {
 	if !cfg.FailOnWarning {
 		t.Error("FailOnWarning should be true")
 	}
-	if cfg.Timeout != 60*time.Second {
-		t.Errorf("Timeout = %v, want 60s", cfg.Timeout)
+	if cfg.Timeout != 120*time.Second {
+		t.Errorf("Timeout = %v, want 120s", cfg.Timeout)
 	}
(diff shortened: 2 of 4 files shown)


Dicklesworthstone:
Thanks — solid diagnosis. Per project policy I don't merge outside PRs, but I reviewed your approach and handled both parts independently:

- **The stdout pipe teardown race** (`reading output: read |0: file already closed`) was already fixed on `main` in `2e7a5a2f` using the same approach you took: drop `StdoutPipe()` + manual `io.ReadAll`, attach a bounded `cmd.Stdout` writer, and use `cmd.Run()` so `os/exec` drains stdout in its own goroutine and waits only after the copy completes. Your `limitedOutputBuffer` and the in-tree `boundedBuffer` are the same idea (never return an error to exec's copy goroutine; surface the oversize condition afterward).
- **The 60s→120s default timeout bump** I just landed in `fe7ded5d` — your point about real Next.js scans taking 70-80s is correct, and 60s was tripping `ErrTimeout` on legitimate scans.

So everything in your PR is now reflected in `main`. Closing this out — appreciate the thorough write-up and repro.

PR #176, opened 2026-06-07
docs(ntm): add checkpoint recovery cadence

christso:
## Summary

Operator recovery guidance now treats checkpoints as a normal cadence instead of a one-off recovery action. The NTM skill also reminds operators that Agent Mail may live outside tmux, so a simultaneous tmux and Agent Mail outage should be investigated at the service/process boundary before workers are relaunched.

## Verification

- `git diff --check -- SKILL.md references/COMMANDS.md`
- `rg -n '[[:blank:]]+$|\t$' SKILL.md references/COMMANDS.md` returned no matches
- fenced-code balance check with `awk` for `SKILL.md` and `references/COMMANDS.md`

`ubs` and `markdownlint` were not available in this environment.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
GPT_5

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -146,6 +146,27 @@ ntm changes conflicts myproject
 ntm resume myproject
 ```
 
+Checkpoint as an operator cadence, not only as a last-resort recovery tool:
+
+```bash
+ntm checkpoint save myproject -m "after spawn and prompt receipt"
+ntm checkpoint save myproject -m "after restore and prompt receipt"
+ntm checkpoint save myproject -m "after investigation: root cause isolated"
+ntm checkpoint save myproject -m "before risky edits: auth refactor"
+ntm checkpoint save myproject -m "after uncommitted edits before verification"
+ntm checkpoint save myproject -m "after green verification"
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks — this is good operational guidance and accurate (`ntm checkpoint save` is a real command). Per project policy I don't merge outside PRs, but I folded the substance into `SKILL.md` independently in `30a8f7d2`: checkpoints framed as a routine cadence (after spawn/restore prompt receipt, after isolating a root cause, before risky edits, after uncommitted work pre-verification, after green verification, before merge/cleanup/handoff), plus the note that Agent Mail can run outside tmux so a joint outage should be diagnosed at the service/process boundary before relaunching workers. Credited the idea in the commit. Appreciate it!

christso:
Created follow-up issue #177 to capture the reusable user story, reproduction scenario, and acceptance criteria for validating the checkpoint/Agent Mail recovery workflow beyond the original incident context: https://github.com/Dicklesworthstone/ntm/issues/177

PR #181, opened 2026-06-09
fix(agentmail): classify lock-list resource timeouts

JYeswak:
## Summary
- map JSON-RPC timeout messages to Agent Mail ErrTimeout
- add typed locks list error_code/reason_code and Agent Mail diagnostics for resources/read failures
- cover the resources/read timeout path without waiting on a real timeout

## Validation
- go test ./internal/agentmail ./internal/cli -run 'TestReadResourceJSONRPCTimeoutMapsToTimeout|TestRunLocksJSONClassifiesResourcesReadTimeout|TestBuildLocksAdviceResult|TestLocksAgentMail|TestRunLocks|TestLocksCheckPathMatches|TestLocksComparable'

SkillOS bead: skillos-ok7k0

### DIFF
--- a/internal/agentmail/errors.go
+++ b/internal/agentmail/errors.go
@@ -152,6 +152,8 @@ func mapJSONRPCError(rpcErr *JSONRPCError) error {
 		return fmt.Errorf("%w: %s", ErrContactBlocked, rpcErr.Message)
 	case strings.Contains(msg, "conflict") && strings.Contains(msg, "reservation"):
 		return fmt.Errorf("%w: %s", ErrReservationConflict, rpcErr.Message)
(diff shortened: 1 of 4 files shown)


Dicklesworthstone:
Good catch — and per policy I don't merge outside PRs, but the diagnosis is correct: in `mapJSONRPCError`, a JSON-RPC `-32000` "request timed out" falls through to the default arm and returns a raw error, so `errors.Is(err, ErrTimeout)` is false and `ntm locks list --json` can't distinguish a timeout from any other failure.

The fix wires into existing infrastructure cleanly (`ErrTimeout`/`IsTimeout` already exist), and the structured `locks list --json` diagnostics (reason_code + a bounded HealthCheck probe for daemon reachability) are a nice agent-ergonomics win. I'll implement equivalent timeout classification + diagnostics independently. Thanks!

PR #188, opened 2026-06-14
Harden agent lifecycle: menu-state recovery, idle-dispatch race, claude --effort threading, kill orphan-reap

code-of-kai:
Four small, independent fixes to the agent lifecycle, each surfaced while running
`ntm assign --watch` autonomous dispatch against a live multi-agent swarm. Together they
close a failure chain that silently stalls a swarm, plus a dispatch race and a process
leak. **Every commit is atomic and independently cherry-pickable** — happy to split into
separate PRs if you'd prefer.

`+408 / −4` across 12 files; a new test accompanies every fix; fixes 1, 2 and 4 are verified
live end-to-end.

### The failure chain (why these belong together)

> spawn an agent at the wrong reasoning effort (fix 1) → operators reach for a post-spawn
> interactive `/effort`, which opens a picker menu → that menu state is invisible to the
> detector, so the agent is stranded and undispatchable (fix 2) → meanwhile auto-dispatch
> can double-assign a pane that's merely between turns (fix 3), and `ntm kill` leaks the
> agent processes on teardown (fix 4).

---

### 1 — Thread reasoning effort into the Claude command  (`f15d685`)  *(the root cause)*

`ntm spawn --cc=N:model:effort` parses the effort into the `AgentSpec` and passes it to the
command renderer, but the default **Claude** template (`internal/config/templates.go`) never
references `.ReasoningEffort` — while the **Codex** template threads `model_reasoning_effort`.
So Claude agents always launch at the saved CLI default, regardless of the requested effort.

`claude --effort <level>` is a real CLI flag; the template simply omitted it. One-line fix,
mirroring Codex:

```gotmpl
{{if .ReasoningEffort}} --effort {{shellQuote .ReasoningEffort}}{{end}}
```

Renders e.g. `claude --dangerously-skip-permissions --model claude-opus-4-8 --effort xhigh`;
omits the flag when effort is empty. Removes the need for the fragile post-spawn interactive
`/effort` workaround that causes the menu strand in fix 2.

*Follow-up (not in this PR): other hardcoded `claude --dangerously-skip-permissions` strings
in `internal/cli/setup.go` and `session_persist.go` may want the same threading.*

### 2 — Recognize + auto-recover the interactive-menu state  (`c81c809`)

A Claude Code pane parked at a selection menu (the `/effort` or `/model` picker, a permission
selector) classified as `UNKNOWN` — neither dispatchable nor recoverable. An agent stranded
there is invisible to `ntm assign`, which reports "0 idle agents" while agents sit
available-but-dead.

- New `AgentState.IsMenuBlocked` + `RECOVER_MENU` recommendation, surfaced in
  `--robot-is-working` as `is_menu_blocked`.
- `detectInteractiveMenu()` keys on the menu footer (a confirm/navigate hint **and**
  "Esc to cancel") within a bounded, newline-tolerant window of the live tail, so prose that
  merely mentions a phrase can't false-match. Covers the `/effort` list-picker footer
  (`Enter to select · Tab/Arrow keys to navigate · Esc to cancel`) and the `/model` footer
  (`Enter to set as default · s to use this session only · Esc to cancel`).
- `tmux.SendEscape()` sends a real **Esc key** — the existing `SendKeys` uses `send-keys -l`
  (literal), which would type the word "Escape".
- `recoverIfMenuBlocked()` Esc-recovers a strand to a clean prompt then dispatches; wired
  into all five dispatch / idle-scan paths in `assign.go`.

Verified live: open a `/model` menu → `menu_blocked=true, RECOVER_MENU` → `ntm assign --pane`
auto-recovers and dispatches.

### 3 — Exclude in-flight-assignment panes from the idle pool  (`c5fbb03`)

`ntm assign --watch` (and one-shot `--auto`) built the idle-agent pool purely from on-screen
state. A pane idle **between turns** of a bead it had already claimed was treated as available
and got a *second* bead piled on (observed: the watch dispatched a new bead to a pane already
working one).

`loadActiveAssignmentPanes()` reads the assignment store's `ListActive()`; the three
idle-collection paths now gate on `state == "idle" && !holdsActiveAssignment`. Safe against
stranding — the watch's completion detector marks an assignment completed when its bead
closes, so the pane is released the moment its work is genuinely done.

### 4 — Reap orphaned agent process trees on `ntm kill`  (`fd9f19f`)

`ntm kill` ran `tmux kill-session`, which terminates the pane **shells** but not their
detached children — node-based agents launched with `--dangerously-skip-permissions` survive
SIGHUP and reparent to init, leaking processes that hold agent-mail registrations and file
locks (observed: several `claude` processes survived a kill, needing manual `kill -9`).

- `collectProcessSubtree()`: **before** `kill-session`, walk each pane's full process tree via
  the existing `internal/process.GetChildPIDs` (the tree must be captured first — after the
  shells exit the children reparent and can't be found from the shell PID). Depth bound +
  per-level cap guard against runaway recursion.
- `reapOrphanedAgents()`: **after** `kill-session`, SIGTERM the captured trees → 750 ms grace →
  SIGKILL survivors (via `process.IsAlive`). Excludes `self`/parent/`pid<=1`, de-dupes, and
  prints `Reaped N orphaned agent process(es)`.

Cross-platform via the existing `internal/process` helpers (`/proc` on Linux, `pgrep`
fallback). Verified live: `ntm kill` now reports `Reaped 2 orphaned agent process(es)` and the
agent PID is gone (previously leaked); unrelated sessions are untouched (trees are rooted only
at the killed session's panes).

---

### Testing

- New unit tests per fix:
  - `internal/agent/menu_detect_test.go` — both real menu footers + wrapped-hint variant
    detected; clean-idle and prose **not** flagged (no false positives).
  - `internal/cli/assign_active_panes_test.go` — active pane included, no-assignment excluded,
    empty store → empty set.
  - `internal/config/claude_effort_test.go` — `--effort` present when set, absent when empty.
  - `internal/cli/kill_reap_test.go` — subtree collection + the reaper SIGTERM/SIGKILLing a
    **real** child process and leaving it dead; `pid<=1`/self exclusion.
- Existing `agent`, `robot`, `config`, and `cli` (kill / lifecycle / assign) suites pass.
- Fixes 1, 2, 4 verified live end-to-end. Fix 3 verified by unit + regression (the exact
  timing window — active assignment + idle screen — is hard to force live without disrupting
  a running swarm).

### Notes for review

- Each commit is atomic and self-contained; cherry-pick or split freely.
- Fix 4 reuses your existing `internal/process` package rather than adding a dependency.
- No changes to public CLI surface beyond an added `is_menu_blocked` field / `RECOVER_MENU`
  recommendation in `--robot-is-working` output, and the new `Reaped N …` line on `ntm kill`.

---

### DIFF
--- a/internal/agent/menu_detect_test.go
+++ b/internal/agent/menu_detect_test.go
@@ -1,0 +1,50 @@
+package agent
+import "testing"
+
+// realMenuFooters are footers captured live from Claude Code pickers. Each must
+// classify the pane as menu-blocked so autonomous dispatch can Esc-recover it.
+var realMenuFooters = map[string]string{
+	"effort/list picker": "  5. Chat about this\n\nEnter to select · Tab/Arrow keys to navigate · Esc to cancel\n",
+	"wrapped nav hint":   "  5. Chat about this\n\nEnter to select · Tab/Arrow keys to\nnavigate · Esc to cancel\n",
+	"/model picker":      "  ◐ Medium effort ←/→ to adjust\n  Enter to set as default · s to use\n  this session only · Esc to cancel\n",
+}
+
+func TestInteractiveMenuDetection(t *testing.T) {
+	for name, footer := range realMenuFooters {
+		body := "  Some agent output above.\n────────────────────────────────────────\n" + footer
+		st, _ := NewParser().ParseWithHint(body, AgentTypeClaudeCode)
+		if !st.IsMenuBlocked {
+			t.Errorf("[%s] IsMenuBlocked=false, want true", name)
+		}
+		if st.IsIdle || st.IsWorking {
+			t.Errorf("[%s] should be neither idle nor working; idle=%v working=%v", name, st.IsIdle, st.IsWorking)
+		}
+		if got := st.GetRecommendation(); got != RecommendRecoverMenu {
+			t.Errorf("[%s] recommendation=%v, want RECOVER_MENU", name, got)
+		}
+	}
+}
+
+func TestCleanIdleNotMenu(t *testing.T) {
+	idle := "✻ Baked for 16m 20s\n\n────────\n❯ \n────────\n  ⏵⏵ bypass permissions on          ·\n"
+	st, _ := NewParser().ParseWithHint(idle, AgentTypeClaudeCode)
+	if st.IsMenuBlocked {
+		t.Errorf("clean idle pane wrongly flagged IsMenuBlocked")
+	}
+	if !st.IsIdle {
+		t.Errorf("clean idle pane should be idle; got idle=%v", st.IsIdle)
+	}
+}
+
+func TestProseNotMenu(t *testing.T) {
+	// Prose mentioning one phrase but not the menu structure must not match.
+	for _, prose := range []string{
+		"I will let you select the option and press Enter when ready.\n❯ \n",
+		"Press Esc to cancel the operation if needed; otherwise continue.\n❯ \n",
+	} {
+		st, _ := NewParser().ParseWithHint(prose, AgentTypeClaudeCode)
+		if st.IsMenuBlocked {
+			t.Errorf("prose wrongly flagged as menu: %q", prose)
+		}
+	}
+}
--- a/internal/agent/parser.go
+++ b/internal/agent/parser.go
@@ -214,6 +214,21 @@ func (p *parserImpl) detectStateFlags(output string, state *AgentState) {
 		return
 	}
 
+	// Interactive-menu detection (checked before idle/working). When a Claude
+	// Code selection menu is the LIVE state — the /effort or /model picker, a
+	// permission selector — the pane is parked awaiting an Enter/arrow choice.
+	// It is neither a dispatchable text prompt (a sent prompt would be typed
+	// into the menu) nor active work, so we surface it as a distinct, recoverable
+	// state instead of letting it fall through to UNKNOWN. Autonomous dispatch
+	// can then Esc-recover the pane to a clean prompt before assigning.
+	if p.detectInteractiveMenu(output, state.Type) {
+		state.IsMenuBlocked = true
+		state.IsIdle = false
+		state.IsWorking = false
+		state.IsInError = p.detectError(output, state.Type)
+		return
+	}
+
 	// Idle detection
 	// We check this BEFORE trusting IsWorking, because IsWorking patterns
 	// (like "testing", "running") might still be present in the scrollback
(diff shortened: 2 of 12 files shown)


Dicklesworthstone:
Thank you for this — the failure-chain framing (wrong effort → interactive `/effort` → menu strand → idle-dispatch race → kill leak) was a genuinely useful way to see the whole lifecycle, and your live verification made it easy to confirm.

Per project policy I don't merge outside PRs, but I reviewed each fix against current `main`, confirmed all four are real, and reimplemented them independently in **`e250fd5a`**:

- **Fix 1 (Claude `--effort` threading)** — confirmed the default Claude template dropped `.ReasoningEffort` while Codex threaded it; `claude --effort <level>` is a real flag. Reimplemented the template threading.
- **Fix 3 (active-assignment idle-dispatch race)** — confirmed idle collection deduped by bead, not by pane. Reimplemented as `loadActiveAssignmentPanes` (via `store.ListActive()` → Assigned|Working) excluded from every idle-collection path.
- **Fix 4 (`ntm kill` orphan reap)** — confirmed `runKill` never reaped the agent subtrees, leaking node/bun agents that hold mail/file-lock state. Reimplemented as a depth-bounded descendant walk + SIGTERM→750ms→SIGKILL after `KillSession`, with self/init exclusion.
- **Fix 2 (menu-blocked state)** — confirmed the gap is real (a `/effort` picker reads as UNKNOWN and strands the pane). Deferred for now: the capture corpus has no real picker footer to derive a safe regex from, and a guessed pattern risks false matches on prose/diff text. I'll add it once I can capture a live `/effort` menu.

The idle/working-detection half (your #189) is also reimplemented in the same commit — see that PR. All reimplemented fixes ship with unit tests (incl. a never-false-idle invariant) and pass build/vet/agent/cli/status suites. Really appreciate the atomic, individually-verified breakdown — it made independent reimplementation straightforward.

code-of-kai:
Quick follow-up, and thanks again for reimplementing these in `e250fd5a` — confirmed the Claude `--effort` threading works great on `main`.

One thing for installer-users: the official `install.sh` pulls the latest **release** tarball, and the newest tag (`v1.18.3`) predates `e250fd5a` — so anyone installing or auto-updating via the released binary still gets the old template that drops `--effort` (and likewise misses the menu-recovery from `c81c809`). Any chance of cutting a release off current `main`? Happy to verify the released artifact once it's up.

No rush — just flagging that the fix is effectively main-only for tarball installs right now.

Issue #187, opened 2026-06-11
3 `/vibing-with-ntm` unstick issues


bitjson:
Fable 5:

> **Title:** `bug(robot): --robot-restart-pane respawns a bare shell instead of the agent, then types --restart-prompt into the shell (runs as a command / traps in quote>), reporting success:true prompt_sent:true`
> 
> ## Summary
> 
> `ntm --robot-restart-pane=<session> --panes=N --restart-prompt "<text>"` does not relaunch the agent. It
> respawns the pane's **default command (the login shell)** and then types `--restart-prompt` into that
> shell. The prompt is executed as shell input — a single-line prompt runs as a command; a multi-line prompt
> or one with an apostrophe traps zsh in a `quote>` continuation and is lost. The command reports
> `success:true, prompt_sent:true, process_alive:{N:true}` regardless (the *shell* is alive).
> 
> ## Root cause
> 
> `GetRestartPane` (`internal/robot/restart_pane.go`):
> - `:145` `tmux.RespawnPane(pane.ID, true)` → `respawn-pane -t <pane> -k` with **no command**
>   (`internal/tmux/client.go:381`). tmux respawns the pane's *default/start* command — the shell.
> - In ntm sessions the agent CLI is started by `send-keys` after spawn (`internal/cli/spawn.go:2136`
>   sends the rendered `claude …`/`codex …` command as keystrokes), so the agent is a child the shell
>   launched, **not** the pane's default command. `respawn-pane -k` cannot bring it back, and there is no
>   agent-relaunch step in the restart-pane path.
> - `:191`–`:197` after a 500 ms wait, `sendRestartPrompts(..., SendKeysForAgentDoubleEnter)` types the
>   prompt into the pane — which is now bare zsh.
> - `:155` the liveness check reads `pane_pid` + `HasChildAlive`/`IsAlive` — the shell is alive →
>   `process_alive:true`, so the false success passes liveness.
> 
> ## Reproduction
> 
> **Deterministic — no agent CLI required.** The placeholder process below stands in for an agent launched
> by keystroke; `respawn-pane -k` brings back the shell, not the placeholder, which is the whole bug. Needs
> only a built `ntm` (`go build -o /tmp/ntm ./cmd/ntm`; verified at HEAD `7d53359b`) and `tmux`/`zsh`. Save
> and run; it self-cleans.
> 
> ```bash
> #!/usr/bin/env bash
[...]

PR #189, opened 2026-06-14
Swarm autonomy: fix Claude idle/working detection + periodic ready-work scan (stacked on #188)

code-of-kai:
## Swarm autonomy: correct Claude idle/working detection + add a periodic ready-work scan

Two fixes that, together, are the difference between an autonomous-dispatch loop that **looks** like it's working and one that actually keeps idle agents fed. Found and verified while running `ntm assign --watch` against a live multi-agent Claude swarm that had silently stalled — four agents idle for ~2 hours while every status surface reported them busy and the watch loop dispatched nothing.

> **Stacked on #188.** This branch builds on #188 (commits `c81c809..fd9f19f`) — the new code uses the `IsMenuBlocked` state and the active-assignment idle-pool guard introduced there. The **two commits unique to this PR** are `4ede0d1` and `88631aa`; review them in isolation here: `compare/fd9f19f...fix/claude-idle-detection`. Please merge #188 first (or merge this after it).

`+~480 / −~30` for the two new commits; a test accompanies every fix; the cycle and restart behavior were verified live end-to-end.

---

### 1 — Correct Claude idle/working classification (`4ede0d1`)

**The bug.** Claude Code pins its input box to the **bottom** of the screen at all times, so the `❯` prompt is the lowest line *even during active work* (the spinner renders above the box). Two failures followed:

- **Idle agents were invisible.** After a turn, Claude parks the pane at a `new task? /clear to save … tokens` footer with the input box holding queued text or an `…` placeholder. No detector recognized that — the idle patterns only matched an *empty* `❯` — so idle agents classified as not-idle and the watch never dispatched to them. This is what stalled the swarm.
- **A dangerous false-idle.** The detectors used a "spinner-above-prompt ⇒ idle" ordering. Since the prompt is *always* below the spinner, this misclassified **busy** panes as idle — a pane actively running (`· Processing… (2m 8s)`) read as idle, which would let autonomous dispatch inject a second bead into a working agent.

**The fix.** A single shared helper, `agent.ClaudeActivelyWorking`, reused by every detector:

> A pane is *working* iff an active spinner is the most-recent **dynamic** marker — i.e. no turn-ended marker (the `new task?` hint or a completion line like `✻ Baked for 16m 20s`) appears after the last spinner. A spinner above a turn-ended marker is stale scrollback.

The scan is window-bounded (the live spinner is always just above the box) and the completion-line pattern is deliberately narrow (glyph-led, whole-line, excludes `…`) so it can't match a `thought for 14s` annotation or prose. The classifier is **biased to the safe failure** (false-*working*, never false-idle). Wired into all four state-detection paths — `agent.detectIdle`, `robot.IsLiveBusy` + the activity classifier, `status.DetectIdleFromOutput`/`determineState`, and `get-all-session-text` — plus a status `pane_activity`-recency fix (a live TUI repaints its idle box, which otherwise pins it to WORKING) and a `ClearPaneInput`/`sendDispatchPrompt` guard so a dispatch clears leftover queued text before typing.

This updates one fixture in `TestParser_Parse_Idle_Claude_StaleSpinnerAbovePrompt` (added in #187): its premise — an active-tense spinner above an empty box ⇒ idle — is the exact false-idle this corrects. The fixture is changed to a realistic finished-turn screen (a spinner *with* a completion + `new task?` marker below it).

### 2 — Periodic ready-work scan in the watch loop (`88631aa`)

**The bug.** `WatchLoop.Run` was **purely completion-driven**: its main `select` waited only on completion events, which the detector fires only for beads *this loop dispatched*. There was no periodic re-scan for idle agents + ready work. So if the initial assignment pass dispatched nothing — no idle agents *or* no ready work at startup (exactly the stall condition above) — **no completion event would ever fire and the loop sat inert forever**, even as ready work appeared (a gate unblocks, beads are created, startup-busy agents go idle). The 30s `watch-interval` only polled for *completions*.

**The fix.** A periodic ready-work scan: `WatchLoop` gains `scanOpts`/`scanInterval`, and the main loop a `case <-ticker.C:` that re-runs the same plan/dispatch pass as the initial assignment (`getAssignOutputEnhanced` + `executeAssignmentsEnhanced`) every `watch-interval`. Idempotent by construction — the idle pool already excludes busy panes and panes holding an active assignment (#188), so a re-scan never double-dispatches in-flight work.

---

### Testing & live verification

- New unit tests: `internal/agent/{cc_newtask_idle,claude_activity}_test.go`, `internal/cli/dispatch_idle_repro_test.go`, `internal/robot/livebusy_ordering_test.go`, `internal/status/{newtask_idle,repaint_idle}_test.go` — all reproduced RED-first from real captured screens (working *and* idle), with explicit guards that a genuinely-working pane is never read as idle. Full `agent`/`robot`/`status`/`cli`/`tmux` suites pass.
- **Live, end-to-end:** all five status surfaces (`status`/`activity`/`health`/`get-all-session-text`/`assign`) agree with reality for both idle and working agents. With a controlled dependency chain of throwaway beads, the loop **cycled unattended** — `a → b → c` dispatched sequentially via the periodic scan, each after the prior closed and unblocked the next — and a **freshly-restarted** loop picked up work that became ready while no dispatcher was running.

### Known limitation

A Claude turn that finishes with a **tiny context** renders neither a completion line nor the `new task?` hint — just an empty box and a *frozen* spinner. A single snapshot cannot distinguish a frozen spinner from a live one (telling them apart needs a second timed sample — velocity). Such a pane reads as **false-working** — the safe direction (no double-dispatch) — and does not occur in real work, where a finished turn accumulates context and renders the hint. Velocity-based liveness in the dispatch path is the robust closure; noted as follow-up, not included here.

### DIFF
--- a/internal/agent/cc_composebox_idle_test.go
+++ b/internal/agent/cc_composebox_idle_test.go
@@ -1,0 +1,57 @@
+package agent
+
+import "testing"
+
+// Real captured idle screens (2026-06-14) from a swarm processing the backlog:
+// agents finished short turns in narrow 7-line panes, so the "new task?" hint is
+// not visible and the ❯ box holds queued text or an … ellipsis. Only the
+// compose-box footer ("⏵⏵ bypass permissions …") is present. These must classify
+// idle (no active spinner) so the dispatcher can hand them the next bead;
+// pre-fix they fell through to UNKNOWN and the backlog stalled with idle agents.
+func TestClaudeComposeBoxIdle(t *testing.T) {
+	cases := map[string]string{
+		"queued-cmd-in-box": "" +
+			"───────────────────────────────────────\n" +
+			"❯ br ready\n" +
+			"───────────────────────────────────────\n" +
+			"  ⏵⏵ bypass permissions on         ·\n",
+		"ellipsis-only": "" +
+			"────────────────────────────────────────\n" +
+			"❯ …\n" +
+			" \n" +
+			"────────────────────────────────────────\n" +
+			"  ⏵⏵ bypass permissions on          ·\n",
+		"completed-then-queued": "" +
+			"✻ Baked for 1m 29s\n" +
+			"────────────────────────────────────────\n" +
+			"❯ next bead\n" +
+			"────────────────────────────────────────\n" +
+			"  ⏵⏵ bypass permissions on (shift+tab to cycle) · ← for agents\n",
+	}
+	p := NewParser()
+	for name, screen := range cases {
+		t.Run(name, func(t *testing.T) {
+			st, _ := p.ParseWithHint(screen, AgentTypeClaudeCode)
+			if !st.IsIdle {
+				t.Errorf("compose-box idle screen classified IsIdle=false (want true); IsWorking=%v", st.IsWorking)
+			}
+		})
+	}
+}
+
+// The compose-box footer must NOT override an active spinner: a working pane has
+// the same footer but with a live spinner, and must stay non-idle.
+func TestClaudeComposeBoxWithActiveSpinnerNotIdle(t *testing.T) {
+	working := "" +
+		"  reasoning about the allocator…\n" +
+		"✻ Noodling… (2m 5s · ↓ 6.2k tokens)\n" +
+		"────────────────────────────────────────\n" +
+		"❯ \n" +
+		"────────────────────────────────────────\n" +
+		"  ⏵⏵ bypass permissions on          ·\n"
+	p := NewParser()
+	st, _ := p.ParseWithHint(working, AgentTypeClaudeCode)
+	if st.IsIdle {
+		t.Errorf("active-spinner pane with compose-box footer classified idle; want not-idle (would interrupt working agent)")
+	}
+}
--- a/internal/agent/cc_newtask_idle_test.go
+++ b/internal/agent/cc_newtask_idle_test.go
@@ -1,0 +1,68 @@
+package agent
+
+import "testing"
+
+// Real captured Claude Code idle screens from a live swarm (2026-06-14). After a
+// turn completes, modern Claude Code parks the pane showing its input box plus a
+// "new task? /clear to save <N> tokens" footer hint. The input box's ❯ chevron
+// may be empty (with an … placeholder) or hold queued, unsent text. None of these
+// matched the pre-fix idle patterns (the ❯…$ pattern needs an *empty* chevron;
+// the ^.{0,40}\?$ pattern needs the ? at line-end), so an idle agent classified
+// as neither idle nor working — invisible to autonomous dispatch.
+var ccNewTaskIdleScreens = map[string]string{
+	"queued-text-in-box": "" +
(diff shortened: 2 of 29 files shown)


code-of-kai:
Pushed two follow-up commits (`258714f`) extending this branch, both surfaced while running the swarm against a real backlog:

- **`{AGENT_NAME}` template substitution** — `expandPromptTemplate` now substitutes the target pane's agent identity, so a dispatch template can instruct the agent to commit under `AGENT_NAME=<name>` (needed when a repo's pre-commit hook requires it and the pane shell doesn't export it). Without it, an autonomous dispatch produces work that can't be committed.
- **Compose-box idle recognition** — a Claude pane that finishes a short turn in a narrow tiled pane shows neither an empty `❯` (box holds queued text) nor the `new task?` hint (scrolled off), so it fell through to UNKNOWN and the dispatcher saw 'no idle agents' while agents sat available. The `⏵⏵` compose-box footer is added as an idle marker across the agent/robot/status detectors, gated through `ClaudeActivelyWorking` (active spinner still reads as working) and backstopped by the active-assignment guard.

Both have tests (incl. a no-false-idle regression). Verified end-to-end: the swarm autonomously dispatched, completed, committed, and closed backlog beads.

Dicklesworthstone:
This was the highest-value finding in the pair — thank you. The "Claude pins its input box to the bottom, so the spinner is always *above* the prompt" insight is exactly right, and it explained both the silent swarm stall and the dangerous false-idle.

Per policy I don't merge outside PRs, but I confirmed both fixes against current `main` (including reproducing the false-idle and false-negative-idle directly from the real captures in `internal/cli/outputs/`) and reimplemented them independently in **`e250fd5a`**:

- **Fix 1 (idle/working classification)** — reimplemented as a shared `agent.ClaudeActivelyWorking` helper: a pane is WORKING iff an active spinner is the most-recent dynamic marker (no completion line / "new task?" footer appears after the last spinner), window-bounded to the live tail and biased to false-WORKING so a busy agent can never be mislabeled idle. The completion-line matcher is glyph-led/whole-line/duration-shaped and explicitly rejects a `thought for …` prose annotation and any `…(` spinner decoration. Wired authoritatively into `agent.detectIdle`/`detectStateFlags`, `status.DetectIdleFromOutput`/`determineState`; `robot.IsLiveBusy` already agreed. I also rewrote the pre-existing fixtures that encoded the *wrong* (spinner-below-prompt) layout — that inverted model was the source of the false-idle — and added real-capture + invariant tests.
- **Fix 2 (periodic ready-work scan)** — confirmed `WatchLoop.Run` was purely completion-driven and would sit inert if the initial pass dispatched nothing. Reimplemented as a watch-interval ticker re-running the plan/dispatch pass, made idempotent by the active-pane guard from #188.

All verified green (build/vet/agent/cli/status). The lifecycle fixes from #188 landed in the same commit. Excellent debugging — the spinner-vs-completion ordering model is now the authoritative signal across every detection path.

PR #190, opened 2026-06-14
Swarm reliability: 7 fixes found by adversarial Haiku stress testing (on e250fd5)

code-of-kai:
Seven small, independent fixes to the autonomous swarm lifecycle, each surfaced by
running a Haiku-only agent swarm under adversarial load (dependency chains, 8-wide
fan-outs, sub-interval bead churn, mid-flight agent kills, watch-loop restarts) in
an isolated sandbox. All build on `e250fd5`; each has a test; the result is
verified live (40-bead churn + 8-wide fan-out to full saturation, zero
double-dispatch, restart-survivable, crash-recovering).

### 1 — Completion detector never reloads its store (`0794a9b`)
The completion detector is handed the watch loop's store instance at construction
and never reloads it, but assignments are recorded by a *separate* store instance
in `executeAssignmentsEnhanced`. So the detector never observes anything dispatched
after startup → never marks those beads completed → never releases their panes.
The on-disk store accumulates stale `assigned` records that permanently mark every
pane busy, and **the autonomous loop dies after at most (pane count) dispatches.**
Fix: reload the store at the top of `checkAll`.

### 2 — False completion: a crashed/slow agent's bead marked "done" (`8122bcc`)
`matchCompletionPatterns` matches the **dispatch prompt's own echo** — the prompt
literally says "run `br close <id>`", and the pattern `br\s+close` matches it. A
crashed or slow agent whose pane still shows the prompt was declared complete,
**silently dropping its work** (and, since completed beads are suppressed from
re-dispatch, stranding it). Also, `checkIdle` reported the idle-timeout as a
*success*. Fix: a completion-pattern match must be confirmed by an authoritative
`checkBeadClosed` re-check; an idle/stalled agent whose bead is not closed is
reported FAILED (so the pane is released and the bead reassigned).

### 3 — Accented Claude spinner verbs break turn-ended detection (`0213bd5`)
`claudeCompletionLineRe` used `[A-Z][a-z]+` for the verb, but Claude's whimsical
spinner verbs include accented forms (`✻ Sautéed for 36s`). The accent broke the
match, so the completion line was not recognized as turn-ended; a stale active
spinner higher in the capture then went unsuppressed and `ClaudeActivelyWorking`
returned true — a **false-working that withholds an idle agent from dispatch**
(observed: "Idle Agents: 1" with 8 idle). Fix: `\p{Lu}\p{Ll}+` (capitalized
Unicode word), preserving the anti-prose capitalized-first-letter constraint.

### 4 — Double-dispatch under concurrency (`b490a15`)
With fast (sub-interval) beads and 8 concurrent agents, the same bead got sent to
a second pane at the instant its completion fired. Four causes: send-before-record
(reordered to record-before-send, releasing via `MarkFailed` on send failure); no
recently-completed guard (`loadHandledBeadIDs` suppresses active OR
recently-completed beads, store-based so immune to br lock contention); closed-bead
re-dispatch (pre-send `bv.GetBeadStatus` open re-check); and logs that counted
*planned* not *sent* beads (iterate by index so `PromptSent` persists; log only
`PromptSent` items).

### 5 — Compose-box idle not recognized → swarm won't start (`6f46444`)
A fresh `ntm spawn` showed **"Idle Agents: 0" and the swarm never started**: a
freshly-spawned agent shows its init prompt prefilled in the box (or just the "…"
placeholder) with no completion line / "new task?" hint yet, so the idle patterns
all miss it. The ungated "bypass permissions" match was (rightly) removed because
it fires during work too — but **gated on `!ClaudeActivelyWorking`** (already
checked in `detectIdle`), the `⏵⏵` compose-box footer is a sound idle signal:
TUI alive at its box + no active work = idle.

### 6 — IsLiveBusy false-busy on a stale spinner (`00e2a28`)
`robot.IsLiveBusy` returned true on ANY `CategoryThinking` match in the live
window — including a STALE spinner (`· Thundering… (4s)`) sitting above the
turn-ended completion line (`✻ Churned for 6s`). `determineAgentState` then
overrode the correct idle verdict to "working", so **the dispatcher saw 0 idle
agents after every burst and the swarm stalled mid-run with ready work waiting.**
Fix: for Claude, defer to the ordering-aware `ClaudeActivelyWorking`.

### 7 — Fixture-dependent test skips on a fresh checkout (part of `b490a15`)
`TestClaudeActivelyWorking_RealCaptures` depends on `internal/cli/outputs/`, which
is gitignored (local-only), so it hard-fails in any fresh checkout/CI. Changed to
`t.Skip` on a missing fixture (the logic is also covered by self-contained tests).

### Testing
New tests per fix (`internal/completion/reload_test.go`,
`internal/agent/claude_accented_verb_test.go`, `claude_composebox_test.go`,
`internal/cli/handled_beads_test.go`, `internal/robot/islivebusy_ordering_test.go`,
plus idle-failure/prompt-echo cases). The `agent`, `completion`, `cli`, `status`,
and `tmux` suites pass. Verified live end-to-end on the reconciled binary
(40-bead high-churn + 8-wide fan-out to full 8-way saturation, zero
double-dispatch; watch-loop restart mid-cascade recovers; crashed agents recover).

Two notes on test state, both already true on `e250fd5` before these commits, so
not introduced here:
- `TestNoNewPlaceholders` was red on `e250fd5` (its own `status/patterns.go`
  comment contains the standalone word "placeholder"). The final tidy commit
  rewords that to "ellipsis", so the stub-scanner is now green.
- `internal/robot`'s `TestLatencyRegression_BuildDigest` is a wall-clock latency
  threshold test that fails on this (loaded VPS) host on `e250fd5` as well — it's
  environment/host-speed sensitive, not affected by these changes.

### DIFF
--- a/internal/agent/claude_accented_verb_test.go
+++ b/internal/agent/claude_accented_verb_test.go
@@ -1,0 +1,34 @@
+package agent
+
+import "testing"
+
+// Claude Code's spinner verbs include accented forms (Sautéed, Flambéed). The
+// turn-ended completion-line regex must recognize them; otherwise a stale active
+// spinner above an accented completion line leaves the finished pane falsely
+// "working", withholding an idle agent from dispatch. Found in a Haiku stress run.
+func TestClaudeActivelyWorking_AccentedCompletionVerb(t *testing.T) {
+	idle := "" +
+		"✶ Sketching… (30s)\n" +
+		"* Sketching… (34s)\n" +
+		"✻ Sautéed for 36s\n" +
+		"────────\n❯ next bead\n────────\n  ⏵⏵ bypass permissions on          ·\n"
+	if ClaudeActivelyWorking(idle) {
+		t.Errorf("accented completion line not recognized as turn-ended; pane reads as working (false-working)")
+	}
+	working := "  thinking\n✻ Sautéing… (5s · ↓ 1.2k tokens)\n────────\n❯ \n────────\n  ⏵⏵ bypass permissions on          ·\n"
+	if !ClaudeActivelyWorking(working) {
+		t.Errorf("active accented spinner read as not-working")
+	}
+}
+
+func TestClaudeIsTurnEndedLine_AccentedVerbs(t *testing.T) {
+	for _, ln := range []string{"✻ Sautéed for 36s", "✶ Flambéed for 2m 5s", "✽ Baked for 16m 20s"} {
+		if !claudeIsTurnEndedLine(ln) {
+			t.Errorf("claudeIsTurnEndedLine(%q) = false; want true", ln)
+		}
+	}
+	// Guard: lowercase prose must NOT match (capitalized-verb requirement preserved).
+	if claudeIsTurnEndedLine("  ...we waited for 5m before retry") {
+		t.Errorf("lowercase prose wrongly matched as a completion line")
+	}
+}
--- a/internal/agent/claude_composebox_test.go
+++ b/internal/agent/claude_composebox_test.go
@@ -1,0 +1,36 @@
+package agent
+
+import "testing"
+
+// A freshly-spawned Claude agent (init prompt prefilled in the box, no completion
+// line / "new task?" hint yet) or a box showing only the "…" placeholder must read
+// IDLE so the dispatcher can start the swarm. Upstream's empty-❯ / completion /
+// new-task patterns miss these; the compose-box footer (gated on
+// !ClaudeActivelyWorking) covers them. Found live: a fresh `ntm spawn` showed
+// "Idle Agents: 0" and the swarm never started.
+func TestDetectIdle_ComposeBoxFreshSpawn(t *testing.T) {
+	p := NewParser()
+	cases := map[string]string{
+		"prefilled-init-box": "  Reread AGENTS.md and\n  continue from where\n────────\n❯ Reread AGENTS.md\n────────\n  ⏵⏵ bypass permissions on  ·\n",
+		"ellipsis-placeholder": "────────\n❯ …\n────────\n  ⏵⏵ bypass permissions on  ·\n",
+	}
+	for name, s := range cases {
+		t.Run(name, func(t *testing.T) {
+			st, _ := p.ParseWithHint(s, AgentTypeClaudeCode)
+			if !st.IsIdle {
+				t.Errorf("compose-box state not idle (IsWorking=%v)", st.IsWorking)
+			}
+		})
+	}
+}
+
+// The compose-box footer must NOT override active work: a live spinner with the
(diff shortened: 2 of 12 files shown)


Dicklesworthstone:
Outstanding work — running an adversarial Haiku swarm against `e250fd5` and surfacing seven concrete failures (three of them gaps in that very commit's detection layer) is exactly the kind of testing that makes this robust. Per policy I don't merge outside PRs, but I confirmed all seven against `main` and reimplemented them independently:

- **#7 (placeholder/stub-scanner red)** — fixed first in `257aba1e`. You were right it was already red on `e250fd5`: my idle-detection comments used the standalone word "placeholder" (Claude's `…` box), tripping `TestNoNewPlaceholders`. Reworded to "ellipsis". (For the `TestClaudeActivelyWorking_RealCaptures`/gitignored-`outputs/` concern: my version of that test uses **inline** captures, not file reads, so it's already fresh-checkout-safe. And the `robot` latency-regression test is host-speed-sensitive, unrelated.)

- **#1–#6** — reimplemented in `bef7ffcb`:
  - **#3 accented verbs** — `claudeCompletionLineRe` verb class ASCII→Unicode `\p{Lu}\p{Ll}+` (so `✻ Sautéed for 36s` is recognized as turn-ended; anti-prose capitalized-first-letter preserved).
  - **#5 compose-box idle** — recognize the `⏵⏵` footer as idle, gated by the existing `!ClaudeActivelyWorking` contract (no ungated bypass-permissions match).
  - **#6 IsLiveBusy stale-spinner** — Claude panes now defer to the ordering-aware `agent.ClaudeActivelyWorking` instead of a position-blind CategoryThinking match.
  - **#1 detector store reload** — `checkAll` reloads the assignment store each tick so post-startup dispatches (recorded by the separate `executeAssignmentsEnhanced` instance) are observed; the loop no longer dies after ~(pane count) dispatches.
  - **#2 false completion** — a completion-pattern hit must be confirmed by an authoritative `checkBeadClosed` (defeats the dispatch-prompt `br close` echo); an idle-timeout with the bead still open is reported FAILED (pane released, bead reassigned) rather than a silent success.
  - **#4 double-dispatch** — record-before-send (with `MarkFailed` rollback on send failure), a store-based recently-completed guard (`loadHandledBeadIDs`), a pre-send `bv.GetBeadStatus` open re-check, and logging only beads actually `PromptSent`.

Every fix has a test; `go build`/`vet`/`gofmt` clean and the agent/completion/robot/cli suites pass. Thank you for the high-churn live verification methodology — the spinner-vs-completion ordering model + the record-before-send discipline are exactly right, and catching the accented-verb regex gap was a great find.

PR #195, opened 2026-06-20
fix(add): thread ReasoningEffort into ntm add (was silently dropped, like #188 for spawn)

code-of-kai:
## What

`ntm add <session> --cc=N:model:effort` parses the `:effort` segment into the `AgentSpec` but drops it before launch, so an "xhigh" pane added to a live session silently runs at the CLI default. This is the **same class** as the Claude `--effort` threading you reimplemented in `e250fd5a` (from #188) — but a **different code path**: that fix landed in `spawn.go`; `add.go` has its own render call and never got it.

## Root cause

`internal/cli/add.go` builds `config.AgentTemplateVars` for `GenerateAgentCommand` without a `ReasoningEffort` field:

```go
finalCmd, err := config.GenerateAgentCommand(agentCmd, config.AgentTemplateVars{
    Model:            resolvedModel,
    ModelAlias:       agent.Model,
    // ... no ReasoningEffort ...
})
```

So the Claude template's `{{if .ReasoningEffort}} --effort {{shellQuote .ReasoningEffort}}{{end}}` clause sees an empty value and renders nothing — even though `agent.ReasoningEffort` was parsed from `--cc=…:effort`.

## Fix

Thread `ReasoningEffort` into the render, mirroring `spawn.go` and the adjacent `resolvedModel` persona-override already in `add.go`:

- `resolvedReasoningEffort := agent.ReasoningEffort`
- persona override: `if strings.TrimSpace(p.ReasoningEffort) != "" { resolvedReasoningEffort = p.ReasoningEffort }`
- `ReasoningEffort: resolvedReasoningEffort` in the `AgentTemplateVars`

Plus a regression test (`TestAddThreadsReasoningEffort`) asserting the rendered Claude command carries `--effort 'xhigh'` and that an unset effort leaves no dangling flag.

## Verification

Unit: `go test ./internal/cli/ -run TestAddThreadsReasoningEffort` passes; full `cli` + `config` packages green; `go vet` clean.

Empirical (process argv — the ground truth), same signal both directions:

```
# before fix — added pane has no --effort:
claude --dangerously-skip-permissions --model claude-opus-4-8

# after fix — `ntm add addtest --cc=1:claude-opus-4-8:xhigh`:
claude --dangerously-skip-permissions --model claude-opus-4-8 --effort xhigh
```

Same policy note as #188: feel free to reimplement rather than merge — the diff is small and the test should drop in cleanly.

### DIFF
--- a/internal/cli/add.go
+++ b/internal/cli/add.go
@@ -553,6 +553,7 @@ func runAdd(opts AddOptions) error {
 
 		// Resolve model alias to full model name
 		resolvedModel := ResolveModel(agent.Type, agent.Model)
+		resolvedReasoningEffort := agent.ReasoningEffort
 		modelRequested := strings.TrimSpace(agent.Model) != ""
 
 		// Check if this is a persona agent and prepare system prompt
@@ -573,11 +574,15 @@ func runAdd(opts AddOptions) error {
 				}
 				// For persona agents, resolve the model from the persona config
 				resolvedModel = ResolveModel(agent.Type, p.Model)
+				if strings.TrimSpace(p.ReasoningEffort) != "" {
+					resolvedReasoningEffort = p.ReasoningEffort
+				}
 			}
 		}
 
 		finalCmd, err := config.GenerateAgentCommand(agentCmd, config.AgentTemplateVars{
 			Model:            resolvedModel,
+			ReasoningEffort:  resolvedReasoningEffort,
 			ModelAlias:       agent.Model,
 			ModelRequested:   modelRequested,
 			SessionName:      session,
--- a/internal/cli/add_test.go
+++ b/internal/cli/add_test.go
@@ -39,6 +39,46 @@ func TestNewAddCmd_RegistersOllamaFlag(t *testing.T) {
 	}
 }
 
+// TestAddThreadsReasoningEffort guards against the `add` path dropping the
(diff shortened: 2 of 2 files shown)


Dicklesworthstone:
Thanks very much for this, and for the careful root-cause writeup tracing it back to the spawn-side fix in #188 — your diagnosis was exactly right: `add.go` had its own `GenerateAgentCommand` call site that never got the `ReasoningEffort` field, so the Claude template's `{{if .ReasoningEffort}} --effort ...{{end}}` clause rendered nothing and an added pane silently launched at the CLI default.

Per project policy I don't merge outside PRs, but I reviewed your approach and independently reimplemented the equivalent fix in a0c7e160. It mirrors `spawn.go` the way you suggested: `resolvedReasoningEffort := agent.ReasoningEffort`, a persona override in the PersonaMap branch (`if strings.TrimSpace(p.ReasoningEffort) != "" { ... }`), and `ReasoningEffort: resolvedReasoningEffort` threaded into the `AgentTemplateVars`. I also added a `TestAddThreadsReasoningEffort` regression test that drives the real `ParseAgentSpec -> Flatten -> GenerateAgentCommand` path and asserts `--effort 'xhigh'` renders (with a negative control for the unset case). I confirmed the pre-fix render dropped the flag entirely before landing it.

Appreciate you flagging this one — closing as addressed in a0c7e160.

PR #203, opened 2026-06-30
Add plugin default model resolution

profplum700:
## Summary
- add plugin-aware default model resolution for agent specs that omit an explicit model
- load `agent.defaults.model` from agent plugin TOML and use it in both `spawn` and `add`
- keep explicit `--agent=1:model` values and global config defaults higher precedence than plugin defaults

## Problem
Agent plugins can declare default environment/working-directory/template behavior, but Hermes-style plugin specs without an explicit model were still spawned with an empty model. That made bare plugin invocations such as `--hermes=1` fail registration even when the plugin configuration carried the right model default.

## Fix
This patch adds a general resolver in `internal/cli` that applies precedence in this order:

1. explicit model in the agent spec (`--agent=1:model`)
2. global defaults for the agent type
3. plugin default model from `agent.defaults.model`

The plugin TOML loader now reads the `model` field, and both `spawn` and `add` call the same resolver.

## Precedent / proof
This follows the Opencode default-model precedent from ntm#193: a plugin should be able to provide the default model while still allowing CLI/global config overrides.

Local A/B proof with a Hermes plugin that declares `agent.defaults.model = "google/gemini-2.5-flash"`:

- unpatched v1.18.3 bare `--hermes=1`: `agents_registered=0`, failed because model was empty
- patched side build bare `--hermes=1`: `agents_registered=1`, model resolved to `google/gemini-2.5-flash`

## Tests
- `gofmt -l internal/cli/add.go internal/cli/agent_spec.go internal/cli/agent_spec_test.go internal/cli/spawn.go internal/plugins/agent.go internal/plugins/plugins_test.go` produced no output
- `go test ./internal/plugins ./internal/cli -run 'TestLoadAgentPlugins_LoadsValidTOML|TestResolveModelWithPlugins_DefaultModel|TestSpawnAndAddPluginFlagsUseGlobalConfigFlag' -count=1`
- `go build ./cmd/ntm`

Posted by Codex on behalf of profplum700.

### DIFF
--- a/internal/cli/add.go
+++ b/internal/cli/add.go
@@ -552,7 +552,7 @@ func runAdd(opts AddOptions) error {
 		}
 
 		// Resolve model alias to full model name
-		resolvedModel := ResolveModel(agent.Type, agent.Model)
+		resolvedModel := ResolveModelWithPlugins(agent.Type, agent.Model, opts.PluginMap)
 		modelRequested := strings.TrimSpace(agent.Model) != ""
 		// Reasoning effort comes from the direct spec (`--cc=N:model:effort`)
 		// parsed onto the FlatAgent, and is overridden by the persona below when
@@ -582,7 +582,7 @@ func runAdd(opts AddOptions) error {
 					systemPromptFile = promptFile
 				}
 				// For persona agents, resolve the model from the persona config
-				resolvedModel = ResolveModel(agent.Type, p.Model)
+				resolvedModel = ResolveModelWithPlugins(agent.Type, p.Model, opts.PluginMap)
 			}
 		}
 
--- a/internal/cli/agent_spec.go
+++ b/internal/cli/agent_spec.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/Dicklesworthstone/ntm/internal/persona"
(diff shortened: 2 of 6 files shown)


Dicklesworthstone:
Thank you for this, @profplum700 — the diagnosis is exactly right: a bare plugin spec like `--hermes=1` was resolving to an empty model (GetModelName returns "" for a non-built-in agent type with no alias) and failing Agent Mail registration with agents_registered=0.

A quick policy note: we don't merge outside PRs into ntm, so I'm closing this one rather than merging it. But I did read through your approach carefully and implemented an equivalent fix independently in 4f75a99b (now on main + master). It:

- loads `model` under `[agent.defaults]` in the plugin TOML loader, and
- adds a plugin-aware resolver used consistently by both `spawn` and `add` at every model-resolution site, preserving your precedence exactly: (1) explicit `--agent=1:model`, (2) global config default for the agent type, (3) plugin `[agent.defaults] model`.

Verified the original bug reproduces (ResolveModel(hermes, "") == "") and added a focused test proving explicit > global default > plugin default, plus that a missing/nil plugin map yields empty.

Really appreciate the clear write-up and the A/B repro — it made confirming the fix trivial. Credit for the idea is yours.

PR #211, opened 2026-07-08
feat(plugins): first-class plugin agents in controller, restart, and pane detection

ryan-brosas:
## Summary

Makes **agent-plugin-defined types first-class** across the four paths where previously only built-in agents (claude, codex, gemini, …) were handled. A plugin registered via a TOML file in `~/.config/ntm/agents/<name>.toml` (e.g. a `pi` plugin — the motivating example) can now be used exactly like a built-in agent for:

1. **Controller** — `ntm controller <session> --agent-type=<plugin>` (was: `unknown agent type`).
2. **Restart relaunch command** — a restarted plugin pane keeps the plugin's command template instead of collapsing to a bare alias.
3. **Pane-type detection** — a pane running `<plugin>` now classifies as `type: <plugin>` instead of `"user"`, which unblocks type-based features (restart selection, dashboard).
4. **Restart readiness gate** — plugin agents are recognized as ready (they have no shared content-readiness signature, so a live agent child process is the readiness signal).

This **completes the plugin dispatch that `spawn`/`add` already had** — `controller`/`restart`/`detection` were the remaining gaps. Mirrors the existing `oc`/opencode fallback comment (ntm#193) and follows the precedent of the recently-merged hermes plugin PR (#203).

## Why

Before this change a plugin agent could be spawned/added but:
- could not serve as the controller (`unknown agent type: pi`),
- a pane running it was reported as `type: "user"`, so `--robot-restart-pane` **skipped it entirely** (`restarted: []`), and even when forced it relapsed to a bare shell (no relaunch command).

This makes plugin agents fully usable end-to-end.

## How (minimal, DRY)

Two shared helpers in `internal/plugins`, wired as **fallbacks behind the built-in checks** (built-in behavior is unchanged — only the previously-failing `default`/`unknown` cases change):

- `ResolveAgentCommand(name, dir)` — name/alias → plugin command template (used by controller + restart relaunch).
- `AgentTypeForCommand(command)` — pane command → plugin type, via a `sync.Once`-cached index (used by pane-type detection). Testable core: `buildPluginCmdIndexFrom(dir)` + `matchCommandIndex(index, command)`.

The tmux-layer dispatch (`detectAgentFromCommand`) uses a package-level `var resolvePluginAgentType` (defaults to `plugins.AgentTypeForCommand`) so it is unit-testable without depending on the real cached/default-dir index.

`plugins/defaultAgentsDir()` intentionally duplicates `config.DefaultPath()`'s env resolution (`NTM_CONFIG` / `XDG_CONFIG_HOME` / `$HOME`) because `plugins` cannot import `config` — the cycle is **transitive** (`config → watcher → agentmail → tmux → plugins`), not direct. The ~10 lines of pure env logic are the standard Go cycle-break; documented inline.

## Verification

- `go build ./...`, `go vet`, `gofmt` clean.
- Tests green **with `-race`**: new `TestResolveAgentCommand`, `TestBuildPluginCmdIndexFromAndMatch` (bare/alias/prefix/`/pi-`/`./pi`/case-insensitive + no-false-match on `pine`/`spi-foo`), `TestDetectAgentFromCommandPluginFallback` (injectable resolver; built-ins still win; unknown → user). Existing `detectAgentFromCommand`, controller, restart, and `selectRestartPaneTargets` tests unchanged.
- **Live (5/5 cycles)** on a real `pi` plugin: `ntm spawn --pi=1` → `ntm controller --agent-type=pi` → `ntm --robot-restart-pane` each yielded `success: true`, `failed: []`, `agent_relaunched: true`, and `type: "pi"` persisted across the restart.

## Backwards compatibility

Additive. Built-in agents are unchanged (checked first). Unknown `--agent-type` still errors. No behavior change for users without plugins.

## Known limitations (intentionally out of scope)

- `--type=<plugin>` filter is **not precise** (it is ignored for plugin types; use `--panes=N` for precise targeting). Fixing it cleanly means touching the hot-path `translateAgentTypeForStatus`, tracked as a follow-up.
- `AgentTypeForCommand` index is `sync.Once`-cached for the process lifetime; a plugin added **mid-run** to a long-lived process (e.g. the monitor) won't be picked up until that process restarts. Documented inline; no current caller needs hot-reload.
- `defaultAgentsDir()` honors `NTM_CONFIG` env but not the `--config` CLI flag (the tmux pane-detection layer has no `--config` context — same boundary as the existing built-in detection).
- `agent_counts` has no per-plugin field (cosmetic only; the `type` field is correct).
- Plugin agents run under a wrapper (`bun`/`node`) are not detected (only direct binaries); the existing `detectAgentFromProcessTree` path is built-in-only.

## Note

`TestNormalizeProjectScopedSessionName_RejectsAmbiguousConfiguredProjectPrefix` fails on clean `04877bc` (pre-existing, unrelated to this change — verified by stashing the branch and testing the clean base).

### DIFF
--- a/internal/cli/controller.go
+++ b/internal/cli/controller.go
@@ -14,6 +14,7 @@ import (
 	"github.com/Dicklesworthstone/ntm/internal/config"
 	"github.com/Dicklesworthstone/ntm/internal/kernel"
 	"github.com/Dicklesworthstone/ntm/internal/output"
+	"github.com/Dicklesworthstone/ntm/internal/plugins"
 	"github.com/Dicklesworthstone/ntm/internal/robot"
 	"github.com/Dicklesworthstone/ntm/internal/tmux"
 )
@@ -171,7 +172,7 @@ Custom prompt files support template variables:
 		},
 	}
 
-	cmd.Flags().StringVar(&agentType, "agent-type", "cc", "Agent type: cc, cod, gmi, agy, cursor, windsurf|ws, aider, oc, or ollama")
+	cmd.Flags().StringVar(&agentType, "agent-type", "cc", "Agent type: cc, cod, gmi, agy, cursor, windsurf|ws, aider, oc, or ollama; or any registered agent plugin (e.g. pi, pia)")
 	cmd.Flags().StringVar(&promptFile, "prompt", "", "Custom prompt file (supports template variables)")
 	cmd.Flags().BoolVar(&noPrompt, "no-prompt", false, "Skip sending initial prompt")
 	cmd.ValidArgsFunction = completeSessionArgs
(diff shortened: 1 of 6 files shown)


ryan-brosas:
Apologies for the noise — I see now in CONTRIBUTING.md that outside contributions aren't accepted here. Closing this PR. (If any of the approach is useful to you, please feel free to take whatever you like with no attribution needed.)

Issue #217, opened 2026-07-10
Allow --pane flag to be used with `ntm --robot-send`


etafund:
Right now, `ntm --robot-send` command broadcasts a message to all agents in a given NTM session. It would be useful for robots to have a way to atomically send a message to a single coding agent pane (via `--pane` flag) within a broader NTM session. 

Right now, robots must use the human-oriented workaround of calling `ntm send <session> --pane=<pane> "<message>"`, which doesn't return a JSON object. Something like this -- ``ntm --robot-send=<session> --pane=<pane> --msg="<message>" ` would be preferrable as it has better agent ergonomics and returns an informative JSON object that can help steer the model.

hiSandog:
It would help to define the JSON error contract when `--pane` identifies no pane or more than one pane, so robot callers can distinguish selection failures from send failures. A test could also cover pane names or indexes containing shell-significant characters to confirm the robot path does not inherit the human command’s quoting behavior. Should omitting `--pane` deliberately preserve the current broadcast semantics for backward compatibility?

Dicklesworthstone:
Shipped in [v1.20.0](https://github.com/Dicklesworthstone/ntm/releases/tag/v1.20.0).

`ntm --robot-send=SESSION --pane=N|W.P|%N --msg=...` now selects exactly one physical pane and returns the normal typed robot JSON response. The implementation rejects empty, ambiguous, nonexistent, and conflicting `--pane`/`--panes` selectors before delivery, and the same target is preserved for `--track`.

Coverage includes `TestRobotSendSingularPaneRealTmux` plus built-binary canonical-pane E2E for exact targeting and failure envelopes. The release commit `4fd9b442` passed two independent canonical DSR quality runs (18/18 checks each), including 121 top-level / 578 total primary E2E passes with zero failures or skips.

PR #218, opened 2026-07-11
fix: add --robot-send-pane flag for atomic single-pane robot sends

AruneshDwivedi:
### Summary

Issue #217 asks for robots to target a single agent pane atomically when using `ntm --robot-send`, instead of only the human-oriented `--panes` filter. This adds a `--robot-send-pane` flag that sets the single pane index and feeds the same `SendOptions.Panes` filter the rest of the send path already honors.

### Changes

- Added `robotSendPane` var and registered `--robot-send-pane` on the root command (the bare `--pane` name is already taken by `--robot-history` on the root command and by subcommands, so a scoped name avoids a flag collision).
- In the `robot-send` path, when `--robot-send-pane` is set it populates the pane filter with that single index and errors if combined with `--panes`.
- Builds and the existing `go build ./cmd/ntm` pass; full `go test -short` requires the `tmux` binary which isn't present in this environment.

### Testing
- [x] `go build ./cmd/ntm` passes
- [ ] `go test -short ./...` (blocked: tmux not installed in sandbox)

### DIFF
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -1362,6 +1362,13 @@ Shell Integration:
 			if robotPanes != "" {
 				paneFilter = strings.Split(robotPanes, ",")
 			}
+			if robotSendPane != "" {
+				if robotPanes != "" {
+					fmt.Fprintf(os.Stderr, "Error: --pane and --panes are mutually exclusive\n")
+					os.Exit(1)
+				}
+				paneFilter = []string{robotSendPane}
+			}
 			// Parse exclude list
 			var excludeList []string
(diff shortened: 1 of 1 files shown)


Dicklesworthstone:
Thanks for this — per project policy I don't merge outside PRs, but I reviewed the approach and the underlying need (#217, atomic single-pane robot send) already ships on main: `--robot-send --pane=N|W.P|%N` (mutually exclusive with `--panes`), which targets the same `SendOptions` pane filter your PR did. Closing as already-implemented. Appreciate the contribution!

Issue #228, opened 2026-07-19
dcg pre-send hook: extracted command starting with "-" crashes clap and fails the whole send


IvanWest33:
## Summary

With `[integrations.dcg] enabled = true`, `ntm send` fails hard (rc=1, nothing delivered to any pane) whenever the prompt contains a markdown bullet line that the pre-send heuristic extracts as a likely command. The dcg check never returns a verdict; it crashes on argument parsing, and the error is treated as fatal for the send.

## Environment

- ntm: devel build from main (hit on a 2026-07-02 build; code path unchanged on current main)
- dcg: 0.6.5
- Linux (Ubuntu), also reproduces on macOS

## Chain

1. `extractLikelyCommands` selects any prompt line containing `&&`, `||`, `;`, `|`, `--force`, `--hard`, ` -rf` or ` -fr`. `normalizeCommandLine` strips `$ `, `> ` and `# ` prefixes but not markdown list markers, so a line like `  - run br list && git status` becomes the candidate `- run br list && git status`.
2. `DCGAdapter.CheckCommand` (internal/tools/dcg.go) runs `dcg --robot test --format json <candidate>` without a `--` terminator. The candidate starts with `-`, so clap parses it as a flag and exits 2: `error: unexpected argument '- ' found`.
3. `maybeBlockSendWithDCG` (internal/cli/send.go) returns any adapter error as fatal, so the whole send fails with `dcg test failed: exit status 2: ...` and nothing is delivered.

## Repro

```
$ dcg --robot test --format json "- run br list && git status" ; echo $?
error: unexpected argument '- ' found
2

$ dcg --robot test --format json -- "- run br list && git status" >/dev/null ; echo $?
0
```

End to end: enable the dcg integration and run `ntm send <session> --file prompt.txt` against a codex pane, where prompt.txt contains a bullet line with `&&`. The send returns success:false with the dcg error recorded in history.jsonl, and nothing reaches the pane.

## Impact

This took our autonomous dispatch fleet down for 6 days. A controller loop drives `ntm send --file` into Codex panes; every assignment prompt contains bullet lines, so every dispatch failed, while short health probe sends without bullets kept succeeding. That made it look like pane or agent trouble instead of transport. Workaround was `[integrations.dcg] enabled = false`.

## Proposed fix

An illustrative PR accompanies this issue (per CONTRIBUTING, for your agents to review, not direct merge):

[...]

PR #229, opened 2026-07-19
fix(send): pass dcg pre-send candidates after "--" and strip bullet markers

IvanWest33:
Illustrative fix for #228, filed per CONTRIBUTING.md: I know direct outside PRs are not merged; this is for your agents to review and adapt.

## What

- `internal/tools/dcg.go`: pass the checked command after the `--` terminator in `CheckCommand` and `CheckCommandExtended`, so a candidate starting with `-` is always treated as the positional COMMAND and never parsed as flags. This is the fix dcg's own error tip suggests.
- `internal/cli/send.go`: strip `- ` and `* ` markdown list markers in `normalizeCommandLine`, so the pre-send guard evaluates the command text rather than the markdown decoration.

## Why

See #228: a prompt bullet line like `  - run br list && git status` is extracted as a candidate, keeps its `- ` marker, crashes dcg's clap parser (exit 2), and `maybeBlockSendWithDCG` turns that adapter error into a hard failure of the entire send. In our deployment this silently killed 100% of controller-driven dispatches for 6 days while short health probe sends kept passing.

## Tests

- New `TestCheckCommandLeadingDashCommand` with a clap-emulating fake dcg binary. On the old argv construction it fails with exactly the production error (`dcg test failed: exit status 2: error: unexpected argument ... found`); with `--` it passes.
- Bullet cases added to the `TestExtractLikelyCommands` table.
- `go test ./internal/tools/` passes. `./internal/cli/` passes except the pre-existing `TestFindGitHookPathLinkedWorktree`, which reads the developer's global `core.hooksPath` and fails on any machine where that is set globally; unrelated to this change.

Not touched: the fail-open vs fail-closed policy for adapter errors in `maybeBlockSendWithDCG` (see the issue).

### DIFF
--- a/internal/cli/send.go
+++ b/internal/cli/send.go
@@ -3353,6 +3353,12 @@ func extractLikelyCommands(prompt string) []string {
 
 func normalizeCommandLine(line string) string {
 	trimmed := strings.TrimSpace(line)
+	if strings.HasPrefix(trimmed, "- ") {
+		trimmed = strings.TrimSpace(strings.TrimPrefix(trimmed, "- "))
+	}
+	if strings.HasPrefix(trimmed, "* ") {
+		trimmed = strings.TrimSpace(strings.TrimPrefix(trimmed, "* "))
+	}
 	if strings.HasPrefix(trimmed, "$ ") {
 		trimmed = strings.TrimSpace(strings.TrimPrefix(trimmed, "$ "))
 	}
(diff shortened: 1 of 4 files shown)


Dicklesworthstone:
Exactly the right diagnosis — thank you. Per policy I don't merge outside PRs, but both halves shipped on main (060e6cb3): dcg candidates are passed after a `--` terminator so leading-dash commands stay positional, and `normalizeCommandLine` strips markdown/prompt decoration (bullets, checkboxes, ordered lists, `$`/`>`/`#` prompts) before the guard evaluates the command — broader than the `- `/`* ` case. Closing as already-fixed.

PR #248, opened 2026-08-10
fix(install): suppress noisy curl 404 when probing checksum candidates

### LINKED ISSUE #247: install.sh: misleading curl 404 error printed during a successful install
## Description

Running the installer succeeds, but prints a scary-looking curl error partway through, making a fully successful install look broken:

```
$ curl -fsSL "https://raw.githubusercontent.com/Dicklesworthstone/ntm/main/install.sh" | bash -s -- --easy-mode
==> Installing ntm...
==> Detected platform: linux_amd64
==> Install directory: /home/user/.local/bin
==> Fetching latest version...
==> Installing ntm v1.22.1 for linux_amd64
==> Downloading from https://github.com/Dicklesworthstone/ntm/releases/download/v1.22.1/ntm_1.22.1_linux_amd64.tar.gz...
==> Verifying checksum...
curl: (22) The requested URL returned error: 404
tar: Ignoring unknown extended header keyword 'LIBARCHIVE.xattr.com.apple.provenance'
==> Installed ntm v1.22.1 to /home/user/.local/bin/ntm
...
==> Installation complete!
```

## Root cause

`download_release_checksums()` in `install.sh` tries three checksum-file candidates in order: `checksums.txt`, `SHA256SUMS`, `<asset>.sha256`. The current v1.22.1 release only publishes `SHA256SUMS` and per-asset `.sha256` files — there is no `checksums.txt` asset, so the first probe always 404s. `download_file()` (used for the probe) always calls `curl -fsSL` without suppressing stderr, so curl's own `curl: (22) The requested URL returned error: 404` line leaks straight to the user's terminal even though the function immediately falls through to the next candidate and succeeds.

The install is not actually broken — checksum verification does complete successfully against `SHA256SUMS` — but the visible 404 makes it look like something failed, which is confusing (I initially assumed the release was missing an asset).

## Expected behavior

Probing for an optional/alternate checksum filename that's expected not to always exist shouldn't print a raw tool error to the user when a later candidate succeeds.

## Fix

I've opened a PR that suppresses curl/wget's own stderr output specifically for the checksum-candidate probing loop (via a `quiet` flag on `download_file()`), while leaving error output intact for the primary asset download itself. Verified locally that the noisy 404 line disappears and checksum verification + install still succeed.

d3v3l0:
## Summary

- Fixes #247: `install.sh` printed a raw `curl: (22) The requested URL returned error: 404` while probing for an optional `checksums.txt` asset that doesn't exist in current releases (only `SHA256SUMS` and per-asset `.sha256` files are published), even though it immediately falls through to the next candidate and the install completes successfully.
- Adds a `quiet` flag to `download_file()` and passes it from `download_release_checksums()`'s probing loop, so expected-to-fail candidate probes don't leak the underlying curl/wget error to the user. The primary asset download still shows errors normally.

## Test plan

- [x] Ran `install.sh` locally against a scratch install dir before the fix — reproduced the `curl: (22) ... 404` line despite a fully successful install.
- [x] Ran it again after the fix — the 404 line is gone, `Verifying checksum...` is immediately followed by the tar extraction, and the install still succeeds.
- [x] Ran the installed binary (`ntm version`) to confirm the extracted binary is valid and checksum verification actually ran (not skipped).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -172,15 +172,26 @@ normalize_release_version() {
     printf 'v%s\n' "${version#v}"
 }
 
-# Download a file
+# Download a file. Pass "true" as the third argument to suppress the
+# underlying tool's own error output (used when a 404 is an expected,
+# silently-handled fallback rather than a real failure).
 download_file() {
     local url="$1"
     local dest="$2"
+    local quiet="${3:-false}"
 
     if has_cmd curl; then
-        curl -fsSL "$url" -o "$dest" || return 1
+        if [ "$quiet" = true ]; then
+            curl -fsSL "$url" -o "$dest" 2>/dev/null || return 1
+        else
+            curl -fsSL "$url" -o "$dest" || return 1
+        fi
     elif has_cmd wget; then
-        wget -q "$url" -O "$dest" || return 1
+        if [ "$quiet" = true ]; then
+            wget -q "$url" -O "$dest" 2>/dev/null || return 1
+        else
+            wget -q "$url" -O "$dest" || return 1
+        fi
     else
         print_error "Neither curl nor wget found. Please install one."
         return 1
@@ -316,7 +327,7 @@ download_release_checksums() {
     for candidate in "${candidates[@]}"; do
         checksum_url=$(build_download_url "$version" "$candidate")
         checksum_path="${dest_dir}/${candidate}"
-        if download_file "$checksum_url" "$checksum_path"; then
+        if download_file "$checksum_url" "$checksum_path" true; then
             printf '%s\n' "$checksum_path"
             return 0
         fi


Dicklesworthstone:
Thanks for the report and the illustrative PR! Per project policy (see the README's *About Contributions* section) I don't merge outside PRs, but your diagnosis was reviewed and was spot-on.

I implemented the fix independently in d9d114e1 with a slightly different shape: instead of a `quiet` parameter on `download_file`, there's a separate `try_download_file` helper used only by the probing loop (keeps the mandatory-download path's error semantics untouched and self-documenting), and the candidate order was flipped to probe `SHA256SUMS` first since that's what current releases actually publish — so the common case never 404s at all. The installer E2E test matrix was updated to pin the new behavior.

Appreciate you taking the time — the issue write-up (#247) made this a five-minute verify-and-fix.

Issue #249, opened 2026-08-10
bug: CM integration silently drops context due to HTTP and CLI schema mismatch


profplum700:
## Summary

NTM 1.22.1's CM clients do not match the HTTP or CLI contracts exposed by cass-memory 0.2.13. Context requests therefore appear successful but deserialize to empty/null fields, so `ntm memory context` and spawn recovery silently omit available CM/CASS history.

## Versions

- NTM: `1.22.1` (`1110bce247d4bd90247301bbea8a01915159c86b`)
- cass-memory / `cm`: `0.2.13`

Both are the latest GitHub releases as of 2026-08-10.

## HTTP mismatch

`internal/cm.Client.GetContext` currently posts a legacy body to `/context` and decodes the HTTP response directly into `ContextResult`:

```json
{"task":"test task"}
```

The current CM server treats POST bodies as MCP JSON-RPC. The legacy request returns HTTP 200 with:

```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32601,"message":"Unsupported method: undefined"}}
```

Because the HTTP status is 200 and unknown JSON fields are ignored, NTM returns a `ContextResult` whose slices are nil rather than surfacing the JSON-RPC error.

Reproduction with an NTM-managed `cm serve --port 8200` daemon:

```bash
ntm memory context "test task" --json
```

Observed:

```json
{
  "relevantBullets": null,
  "antiPatterns": null,
  "historySnippets": null,
  "suggestedCassQueries": null
}
```

Direct CM MCP `tools/call` for `cm_context` returns populated `content[].text`, including history snippets.

## CLI recovery mismatch

`CLIClient.GetContext` unmarshals `cm context ... --json` directly into `CLIContextResponse`. CM 0.2.13 returns its data under the standard success envelope:

```json
{
  "success": true,
  "command": "context",
  "data": {
    "relevantBullets": [],
    "antiPatterns": [],
    "historySnippets": [],
    "suggestedCassQueries": []
  },
  "metadata": {},
  "timestamp": "..."
}
```

NTM reads `success` but misses every field under `data`. This affects the CLI client used by spawn recovery.

## Related endpoints

The same client also uses legacy `GET /health` and `POST /outcome` requests. CM's current server is MCP HTTP-only (`ping` and `tools/call`, including `cm_outcome`), so those paths should be reviewed at the same time. In particular, an HTTP 200 JSON-RPC error must not be treated as a successful outcome write.

## Expected behavior

[...]

PR #250, opened 2026-08-11
fix(cm): use current MCP server contract

profplum700:
## Problem

NTM recovery still calls the removed CM REST endpoints (`/context`, `/health`, and `/outcome`). Current CM serves JSON-RPC MCP at its root, so recovery emits timeout/failure warnings even while both CM and CASS are healthy. The current CM CLI also wraps context under `data`, and CASS history snippets now carry source metadata instead of placeholder `id`/`content` fields.

Refs #249.

## Changes

- route CM health through MCP `ping`
- route context and outcome through MCP `tools/call` (`cm_context` and `cm_outcome`)
- retain the NTM session ID required by `cm_outcome`
- preserve legacy sentiment in outcome notes because the current MCP schema has no sentiment field
- decode both the current CLI envelope and the legacy bare response
- preserve full CASS history metadata in the serve adapter
- update focused HTTP adapter and CM tests for the current wire contract

## Validation

- `go test -short ./internal/cm ./internal/tools` via RCH: pass
- `go test -short ./internal/serve -run TestHandleMemoryContext` via RCH: pass
- `go build ./cmd/ntm` and release build via RCH: pass
- live `ntm memory context` against CM: 10 history snippets, no degraded state
- isolated recovery-enabled spawn: recovery prepared without CM, Agent Mail, or recovery-window warnings; Codex reached a working prompt and Agent Mail connected

The repository-wide `go test -short ./...` was also run through RCH. The changed CM and serve packages passed; the run still has unrelated timing/environment failures in existing e2e pipeline, Agent Mail call-count, approval/audit timers, scheduler cooldown, tmux multibyte, and TUI color tests. `make lint` could not execute `golangci-lint` because v2.12.2 is not installed on the worker.

### DIFF
--- a/internal/cm/client.go
+++ b/internal/cm/client.go
@@ -20,8 +20,27 @@ import (
 var ErrNotInstalled = fmt.Errorf("cm is not installed")
 
 type Client struct {
-	baseURL string
-	client  *http.Client
+	baseURL   string
+	sessionID string
+	client    *http.Client
+}
+
+type mcpRPCError struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+}
+
+type mcpRPCResponse struct {
+	Result json.RawMessage `json:"result"`
+	Error  *mcpRPCError    `json:"error"`
+}
+
+type mcpToolResult struct {
+	Content []struct {
+		Type string `json:"type"`
+		Text string `json:"text"`
+	} `json:"content"`
+	IsError bool `json:"isError"`
 }
 
 // PIDFileInfo matches supervisor.PIDFileInfo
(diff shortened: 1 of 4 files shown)


Dicklesworthstone:
Thanks for the PR — per project policy (README, *About Contributions*) I don't merge outside PRs, but I reviewed your approach closely and it correctly identified the whole shape of the problem: MCP JSON-RPC at the root, `tools/call` for context/outcome, session ID for `cm_outcome`, sentiment folded into notes, and the CLI `{success,data}` envelope.

I implemented the fix independently in 6935c2d8 (hardened in b3dc8346). A few places where my version deliberately diverges from this diff, for the record:

- **Health probes `tools/list` rather than `ping`** — the live 0.2.x daemon I tested answers `ping` with `-32601 Unsupported method`, while `tools/list` works on every MCP-speaking release.
- **Envelope detection keys off member presence, not content length** — with `Content []struct` checked via `len(...) > 0`, an envelope with empty content and `isError:false` would fall through to direct-object decoding and silently produce empty context (the exact #249 failure mode resurrected). Presence-based detection makes that case a loud error.
- **`rpc` requires a JSON-RPC result even when the caller discards it**, so Health can't bless an arbitrary HTTP responder on a reused port.
- The legacy snippet/rule types were unified into one schema (real CASS-hit fields) rather than aliasing `CLIHistorySnip` into the HTTP result, and NTM's `partial` outcome maps to CM's `mixed` enum.

Appreciate the careful work — between the issue and this diff, verification was fast and the final fix is stronger for it.

PR #264, opened 2026-08-25
fix(testutil): trust TMUX_TMPDIR, not a flag, for tmux test isolation bypass

gabrielassisxyz:
## What

`NTM_TEST_TMUX_ENV_OWNED=1` let a test binary skip creating its own isolated tmux root, on the assumption that whatever process set the flag already owns one. `IsolateTmuxTestProcess` (`tests/testutil/throttle.go`) and a hand-copied duplicate in `internal/tmux`'s own `TestMain` both trusted the flag alone. Between them, one exported environment variable disabled isolation for `internal/cli`, `internal/checkpoint`, `internal/robot`, `internal/tmux`, `e2e` and `tests/e2e` at once — and with isolation off and no `TMUX_TMPDIR` set, tmux resolves its default socket, so every `kill-server` and `KillSession` in those packages targets whatever tmux server the calling shell happened to be using.

Both call sites now require proof, not just the flag: `TMUX_TMPDIR` must be non-empty and match this package's own `ntm-tmux-test-*` naming pattern. When the flag is set but that doesn't hold, isolation runs normally and a single line names both the flag and the resolved `TMUX_TMPDIR` on stderr, instead of silently trusting it.

`internal/tmux`'s `TestMain` also had its own narrower directory-creation helper, missing the `NTM_TMUX_TEST_TMPDIR` override and the socket-path length validation the rest of the suite already enforces. Both the ownership check and the validated directory creation now live in one place — a new leaf package, `tests/testutil/tmuxenv` — that both `tests/testutil` and `internal/tmux`'s test files depend on. That package exists specifically because `internal/tmux/session_test.go` cannot import `tests/testutil` directly: `testutil` already imports `internal/tmux` for command-error classification, so a test file inside package `tmux` importing `testutil` would be an import cycle. `tmuxenv` has no dependency on `internal/tmux`, so both sides can share it without one.

The flag itself is kept, not removed: seven re-exec sites (`internal/tmux/session_test.go`, `internal/tmux/extra_test.go`, `internal/cli/assign_test.go`, `internal/cli/scale_test.go`) legitimately rely on it, and each builds its child's environment by extending `os.Environ()`, so the parent's already-isolated `TMUX_TMPDIR` reaches the child by inheritance rather than the child creating a second isolated root. Deleting the flag would make each of those children build their own isolated root inside a process whose parent already has one, and the child's `kill-server` cleanup would race the parent's. Trusting inheritance instead of the bare flag is exactly what keeps those seven sites working unmodified while closing the gap for everyone else.

This is test infrastructure — no production code changes; no user-facing CLI/robot surface is involved.

## Why

An ambient shell that merely kept the flag from an earlier command left no trace in a test failure and no trace in the repo (nothing in `scripts/`, the `Makefile`, or `.github/` sets it), so the previous behavior was silently unsafe rather than loudly broken. The fix ties trust to something the flag-setter cannot fake without actually holding an isolated `TMUX_TMPDIR`: the directory name this package itself chooses when it creates one.

Two directions are covered by new tests in `tests/testutil`: `TestIsolateTmuxTestProcessIgnoresUnprovenOwnershipFlag` (flag set, `TMUX_TMPDIR` not owned → isolation still runs) and the updated `TestIsolateTmuxTestProcessPreservesConfiguredFakeBinaryEnvironment` (flag set, `TMUX_TMPDIR` genuinely owned → no-op as before). `tests/testutil/tmuxenv`'s own tests cover the predicate directly: false for an empty `TMUX_TMPDIR`, false for one outside the pattern, true for one this package created.


gabrielassisxyz:
Opened against the wrong repository by mistake (this branch's actual base is the gabrielassisxyz/ntm fork, which has diverged from this upstream). Closing; re-opened correctly there.

PR #272, opened 2026-08-29
fix(activity): expose window and stable pane identities

anupamchugh:
## Summary

- add `window_index` and stable tmux `pane_id` to `ntm activity --json`
- show `Win` and `PaneID` in the activity table while retaining the existing `Pane` field
- add a real two-window tmux regression where both panes have the same window-local index

In multi-window sessions, pane indices are local to each window. Two panes can therefore both appear as `Pane 1`, making activity rows ambiguous. The added fields make each row identifiable without changing the existing `pane` field.

## Verification

- `gofmt -l internal/cli/activity.go internal/cli/activity_collision_test.go`
- `go vet ./internal/cli`
- `go test ./internal/cli -run TestCollectActivityData_DisambiguatesCollidingPaneIndexAcrossWindows -count=3`

The full `internal/cli` suite has the same pre-existing unrelated failures on clean upstream and this branch; the focused regression is green.

### DIFF
--- a/internal/cli/activity.go
+++ b/internal/cli/activity.go
@@ -240,13 +240,15 @@ type activityResult struct {
 }
 
 type agentInfo struct {
-	Pane       int
-	AgentType  string
-	State      string
-	Confidence float64
-	Velocity   float64
-	Duration   time.Duration
-	StateSince time.Time
+	Pane        int
+	WindowIndex int
+	PaneID      string
+	AgentType   string
+	State       string
+	Confidence  float64
+	Velocity    float64
+	Duration    time.Duration
+	StateSince  time.Time
 }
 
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks @anupamchugh — the diagnosis is exactly right: in a multi-window session every window has its own "pane 1", so the per-window index is neither unique nor stable, and two `ntm activity` rows could be indistinguishable in both the table and `--json`.

Per this repo's policy, outside PRs aren't merged directly, so this was implemented independently in 0419c724: `window_index` and `pane_id` (the stable tmux `%id`) added to the JSON envelope, `Win`/`PaneID` columns added to the table, the legacy `pane` field kept for compatibility, and a regression test covering the two-window collision topology end to end.

Closing with appreciation — the idea landed.

PR #274, opened 2026-08-31
fix: make coordinator identity explicit

lifeofgurpreet:
## Summary

- separate the physical coordinator checkout from its canonical Agent Mail project key
- require an explicit `coordinator bootstrap` step that persists one exact `program=ntm`, `model=coordinator` sender/token and exact live pane bindings
- make status/run/digest/conflict dispatch fail closed when coordinator state is absent; ignore unbound, cross-project, and recycled-PID panes
- launch native controllers only in a new detached window, never into an existing pane with index 1

`auto_register` remains unchanged and disabled by operator policy. Bootstrap never creates pane identities, and `--dry-run` returns before coordinator registration or state writes.

## Focused proof

- `env -u AGENT_MAIL_URL -u AGENT_MAIL_TOKEN go test ./internal/agentmail ./internal/coordinator`
- `env -u AGENT_MAIL_URL -u AGENT_MAIL_TOKEN go test ./internal/cli -run "Controller|Coordinator" -count=1`
- `git diff --check`

The controller regression test models existing `%44` and `%51` panes/PIDs and proves only a newly returned pane ID is accepted. Coordinator tests prove strict sender/token persistence, exact five-pane import, cross-project `%66` rejection, unbound/recycled-PID skipping, and fail-closed loading.

## Operational boundary

No live tmux session, coordinator process, Agent Mail identity, network setting, or file permission was changed while preparing this PR.

### DIFF
--- a/internal/agentmail/coordinator_state.go
+++ b/internal/agentmail/coordinator_state.go
@@ -1,0 +1,155 @@
+package agentmail
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/Dicklesworthstone/ntm/internal/util"
+)
+
+const coordinatorStateVersion = 1
+
+// CoordinatorIdentity is the durable Agent Mail sender used by the NTM
+// coordinator. The registration token is intentionally persisted in the
+// same mode-0600 session artifact as the identity; dispatch never invents a
+// sender or silently registers one.
+type CoordinatorIdentity struct {
+	AgentName         string `json:"agent_name"`
+	RegistrationToken string `json:"registration_token"`
+	Program           string `json:"program"`
+	Model             string `json:"model"`
+}
+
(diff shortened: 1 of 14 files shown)


Dicklesworthstone:
Thanks @lifeofgurpreet — you're right that coordinator identity was genuinely broken: the hard-coded `NTM-Coordinator` sender was never a registered Agent Mail identity, so the server rejected mail and claims made under it.

Per this repo's policy outside PRs aren't merged directly, and the fix landed via a lighter-weight route than a mandatory bootstrap step: in c79cc100 the coordinator now registers (or reuses) a server-generated session identity via `RegisterSessionAgent` at first use — the same identity `ntm spawn` registers — persisting the name and registration token so later processes re-claim it; the legacy literal survives only as the offline last resort. This keeps `coordinator status`/`digest`/`run` working with zero setup while sending under a real registered identity.

Two ideas from this PR are being taken forward separately: the detached-controller-window pattern and the stale-pane-binding validation. Closing with thanks — the diagnosis was on the money.

PR #275, opened 2026-08-31
fix: restore durable Agent Mail assignment flow

luminous-maximus:
## Summary

- let spawn prove Agent Mail availability through ensure_project instead of the slower diagnostic health_check, so healthy loaded servers still produce pane identities and agent_registry.json
- register and reuse the session coordinator server-generated identity instead of the invalid hard-coded NTM-Coordinator sender; persist its registration token
- classify explicit MCP tool rejections as known no-actuation failures so atomic assignment returns from sending to pending, while keeping timeouts outcome-unknown
- support assign --clear --force for intentionally releasing outcome-unknown dispatches
- decode structured pane-binding identity files to their name field instead of using the full JSON object as a Beads assignee

## Verification

- go build ./cmd/ntm
- focused changed-path tests across internal/agentmail, internal/assignment, internal/coordinator, and internal/cli
- go test ./internal/agentmail ./internal/assignment ./internal/coordinator (pass)
- go test -short ./...: all changed packages pass, including internal/cli; one unrelated pre-existing failure in internal/bv TestGetTriage because its fixture reported zero issues
- gofmt -l . clean

The non-short internal/cli suite also reached only unrelated tmux fixture failures where test agent commands exited immediately to bare zsh/sh.

### DIFF
--- a/internal/agentmail/client.go
+++ b/internal/agentmail/client.go
@@ -479,11 +479,11 @@ func extractMCPContent(result json.RawMessage) (json.RawMessage, error) {
 			msgLower := strings.ToLower(errMsg)
 			// Detect transient busy errors so callers can retry
 			if strings.Contains(msgLower, "busy") || strings.Contains(msgLower, "temporarily unavailable") {
-				return nil, fmt.Errorf("%w: %s", ErrTransientBusy, errMsg)
+				return nil, &ToolRejectionError{Err: fmt.Errorf("%w: %s", ErrTransientBusy, errMsg)}
 			}
-			return nil, fmt.Errorf("tool error: %s", errMsg)
+			return nil, &ToolRejectionError{Err: fmt.Errorf("tool error: %s", errMsg)}
 		}
-		return nil, fmt.Errorf("tool returned error")
+		return nil, &ToolRejectionError{Err: errors.New("tool returned error")}
 	}
 
 	// Prefer structuredContent (already parsed JSON)
--- a/internal/agentmail/client_test.go
+++ b/internal/agentmail/client_test.go
@@ -898,6 +898,19 @@ func TestExtractMCPContent(t *testing.T) {
 	}
 }
 
+func TestExtractMCPContentTypesToolRejection(t *testing.T) {
+	_, err := extractMCPContent(json.RawMessage(`{
+		"content": [{"type": "text", "text": "Agent 'NTM-Coordinator' not found"}],
+		"isError": true
+	}`))
+	if err == nil || !IsToolRejection(err) {
+		t.Fatalf("extractMCPContent error=%v, want typed tool rejection", err)
+	}
+	if IsServerUnavailable(err) || IsTimeout(err) {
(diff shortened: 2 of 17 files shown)


Dicklesworthstone:
Thanks @george-a-ata — this was a genuinely useful audit. Per this repo's policy outside PRs aren't merged directly, but two of your five findings have already landed as independent implementations, with credit in the commit message (c79cc100):

- **Structured pane-binding JSON as assignee**: confirmed against the current mcp-agent-mail reference implementation (it writes a `PaneIdentityRecord` JSON object to the canonical identity file). `readIdentityFile` now decodes structured content down to its `name` field and fails closed on malformed/name-less JSON; plain-text legacy files resolve unchanged. Tests cover the structured, malformed, and legacy cases.
- **Registration-token persistence**: `SessionAgentInfo` now carries `registration_token`, `RegisterSessionAgent` persists it and primes the client token cache on the re-claim path, so later processes re-claim the same identity. This also underpins the coordinator-identity fix (the hard-coded `NTM-Coordinator` sender is replaced by a server-registered session identity in the same commit).

The remaining three — the `ensure_project`-as-availability-probe change, the `ToolRejectionError` typed rejection / timeout outcome split, and `--clear --force` for outcome-unknown assignment barriers — are accepted direction and tracked for follow-up; they touch the dispatch crash-recovery semantics and deserve their own careful pass rather than riding along here.

Closing with thanks.

Dicklesworthstone:
Following up: the remaining three accepted-direction fixes from this audit have now landed independently on main in 68ca5ae2179ef52db9683009514f03416be83371 (mirrored to master).

- **`ensure_project` as the availability probe**: `RegisterSessionAgent` and the spawn identity coordinator no longer gate on the `health_check` diagnostic (whose snapshot can be slower than ordinary identity operations on a loaded server). `ensure_project` — the call both paths need anyway — is now the reachability probe, run even on the re-registration fast path in case the server rebuilt its database. A loaded-but-healthy server therefore still yields pane identities. Regression tests assert `health_check` is never called on the registration paths.
- **Typed `ToolRejectionError` / timeout outcome split**: `extractMCPContent` now returns a typed `ToolRejectionError` for `isError` tool results (transient-busy classification preserved via `Unwrap`). The coordinator's atomic dispatch port maps that — and only that — to `GuaranteeNoActuation`, so an explicit server rejection returns the assignment from `sending` back to `pending` (retryable), while timeouts and transport failures remain outcome-unknown, preserving at-most-once delivery.
- **`ntm assign --clear --force` for outcome-unknown dispatches**: `BeginClearForce`/`BeginClearForceIfStatus` let an operator release an assignment stranded at the `dispatch_state="sending"` barrier; the dispatch state is terminalized to `pending` with the reason recorded in `last_dispatch_error`, status guards still apply, and the CLI prints a duplicate-delivery warning before proceeding. Without `--force` the barrier keeps rejecting the clear, exactly as before.

Thanks again for the audit — all five findings are now addressed on main.

PR #276, opened 2026-08-31
robot: opt-in CPU-delta liveness fallback for StateUnknown panes

luminous-maximus:
## Summary

Adds an **opt-in** process-evidence fallback for the one case a screen/title observer cannot resolve: a pane the canonical observer reports `StateUnknown` gets a short `/proc`-style CPU-tick delta sample over its process tree instead of being left `Unknown` outright.

## Motivating case

A working `claude` pane rendered a frozen frame with an empty composer for over an hour while actively producing commits and mail the whole time. The canonical observer had no title/screen signal left to read and correctly reported `StateUnknown` — but the pane's process tree was not idle at all. A lifetime-average `%CPU` reading (`ps`) is too diluted by process age to catch this; a short windowed delta is not.

## What's here

- `internal/process/cpu_delta.go`: `CPUDeltaPercent`/`CPUDeltaWorking` sample a pid's full descendant tree's User+System CPU seconds twice, `interval` apart, via `gopsutil` (already a dependency here via `StartTime`), so this is cross-platform for free.
- `internal/robot/is_working.go`: three new `IsWorkingOptions` fields (`CPUFallbackOnUnknown`, `CPUFallbackInterval`, `CPUFallbackThresholdPct`), **off by default**, following the same opt-in pattern as the existing `Semantic` field. `applyCanonicalWorkSafety`'s `StateUnknown` arm calls the fallback only when enabled and the pane has a live PID.

## Open design question — not resolved by this PR

This function **blocks** for `CPUFallbackInterval` per Unknown pane it fires on, sampled serially, one pane at a time. That's an acceptable cost for a slow, occasional caller (a preflight check run once per session) and likely wrong for a hot per-pane status path called frequently across many panes.

Batching every Unknown pane's sample into one shared sleep is the right shape for that kind of caller — an interim Python implementation of this exact idea does that (one sleep for N panes, not N sleeps) — but that batching is intentionally **not** implemented here. The flag exists so this tradeoff is opt-in and discussed at review, not silently defaulted on for every `GetIsWorking` caller.

Happy to follow up with a batched variant (e.g. a session-level pre-pass that samples all Unknown panes' PIDs together) if that's the preferred shape — wanted to get the core detection logic and the opt-in surface in front of you first rather than guess at the caller-side architecture.

## Testing

- `go build ./...` clean.
- `go vet` clean on both touched packages.
- New tests in `cpu_delta_test.go`: busy process reads ≥50% CPU, idle/sleeping process reads <10%, invalid/nonexistent pid returns an error, non-positive interval falls back to a 1s default.
- All pre-existing `internal/robot` and `internal/process` tests still pass, including the 7 call sites of `applyCanonicalWorkSafety` updated for the new parameter.
- One unrelated pre-existing failure (`TestGetSendIdempotency_PreflightFailureReleasesClaim`, a grok-pane preflight error-code expectation) reproduces identically on unmodified `main` via `git stash` — confirmed not caused by this patch.

## Companion fix

An interim Python implementation (batches all panes into one shared sleep, ships independently of this PR's review) is live at `agent-factory/src/pane_liveness.py`, wired into that fleet's preflight check, for the same underlying bug.

### DIFF
--- a/internal/process/cpu_delta.go
+++ b/internal/process/cpu_delta.go
@@ -1,0 +1,114 @@
+package process
+
+import (
+	"fmt"
+	"time"
+
+	gopsutil "github.com/shirou/gopsutil/v4/process"
+)
+
+// CPUDeltaWorking samples cumulative CPU time (user+system) for pid and its
+// full descendant process tree, twice, `interval` apart, and reports whether
+// the tree's CPU usage over that window is at or above thresholdPct.
+//
+// This exists as a liveness fallback for exactly the case a screen/title
+// observation cannot resolve (ntm-tnh1, dea/agent-factory fleet, 2026-08-30):
+// a working `claude` pane can render a frozen frame with an empty composer
+// for over an hour while genuinely producing commits and mail the whole
+// time -- the canonical observer correctly has no title/screen signal to
+// read and reports StateUnknown, but the pane's process tree is not idle at
+// all. A lifetime-average %CPU reading (e.g. via `ps`) is too diluted by
+// process age to catch this; a short windowed delta is not.
+//
+// Uses gopsutil (already a dependency here, see StartTime above) rather than
+// a hand-rolled /proc reader, so this is cross-platform for free -- Linux,
+// macOS, and Windows all implement Process.Times().
+//
+// Cost: this function blocks for `interval`. It is deliberately NOT wired
+// into any hot per-pane status path in this patch -- see the call site in
+// is_working.go's applyCanonicalWorkSafety for the open question this needs
+// resolved before merge (batch multiple panes' samples into one sleep, the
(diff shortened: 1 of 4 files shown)


Dicklesworthstone:
Thanks @george-a-ata — right diagnosis, and the guardrails are exactly the ones we'd want: opt-in only, and a tree-summed CPU delta rather than a single-process sample (which misses work happening in child processes).

Per this repo's policy outside PRs aren't merged directly, so this will be implemented independently — and specifically with the batched shape you flagged as the open question: one sleep window shared across all N sampled panes (snapshot all trees, sleep once, snapshot again) instead of a per-pane sleep, so the fallback's latency cost is constant in the number of panes rather than linear.

Closing with thanks; the design conversation in this PR is the spec for the follow-up.

PR #278, opened 2026-08-31
fix(coordinator): give Grok panes mail-nudge parity

luminous-maximus:
Adds Grok to `rotationSafetySkipReason` so Grok panes get idle-detection and mail-nudge parity through the existing `GrokActivelyWorking` detector. Unknown agent types still fail closed with `unsupported_agent_type`.

Behavioral coverage proves:
- an idle Grok pane with unread Agent Mail is nudged;
- an in-flight Grok spinner / `Esc:cancel` frame is skipped as `working`;
- text already held in the Grok composer is skipped as `unsubmitted_input`.

Verification:
- `go build ./cmd/ntm`
- `go test ./internal/coordinator -count=1`
- focused Grok mail-nudge tests

Tracking: agent-factory-ssrq.

### DIFF
--- a/internal/coordinator/mail_nudge_test.go
+++ b/internal/coordinator/mail_nudge_test.go
@@ -208,6 +208,76 @@ func TestMailNudgeWorkingPaneSkippedAndPublished(t *testing.T) {
 	}
 }
 
+// TestMailNudgeGrokIdlePaneDispatched proves Grok Build panes use their
+// landed live-state detector instead of being rejected as an unsupported
+// agent type. The empty bordered composer is the idle state observed in a
+// real Grok TUI capture.
+func TestMailNudgeGrokIdlePaneDispatched(t *testing.T) {
+	server := newMailInboxServer(t, map[string][]map[string]interface{}{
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks — this one verified cleanly and the gap was real: GrokActivelyWorking, the Grok composer markers, and Grok rate-limit detection all landed with full capture coverage, yet the shared fire-time gate still refused Grok panes as unsupported_agent_type, so mail nudge never reached them.

Per this repo's policy, outside PRs aren't merged directly; I read the diff as a report and reimplemented it independently on main in fb23c51a: the Grok case in rotationSafetySkipReason (working veto via GrokActivelyWorking; the shared rate-limit/interactive-gate/composer checks already handle Grok), a comment documenting why context rotation itself still only reaches the gate for types with transcript locations, and behavioral tests for the idle-nudge, in-flight-spinner, and held-composer-text cases — the same three behaviors your test plan covered. Your capture strings for the Grok TUI states matched what the detectors expect, which made verification quick. Closing with the work landed; thanks again.

PR #279, opened 2026-08-31
fix: use configured Agent Mail client and guard stale panes

lifeofgurpreet:
## Summary

- inject NTM’s configured Agent Mail client into robot mail checks
- report Agent Mail URL/token unavailability as a configuration failure, not a missing dependency
- fail closed on stale tmux pane IDs before restart key injection

## Tests

- `env -u AGENT_MAIL_URL -u AGENT_MAIL_TOKEN go test ./internal/robot -run "^TestGetMailCheckUsesInjectedClientForConfiguredEndpointAndToken$" -count=1`
- `env -u AGENT_MAIL_URL -u AGENT_MAIL_TOKEN go test ./internal/resilience -run "^TestRestartAgentRetiresStalePaneWithoutSendingKeys$" -count=1`
- `env -u AGENT_MAIL_URL -u AGENT_MAIL_TOKEN go test ./internal/agentmail -run "^(TestNewClient|TestBaseURL)$" -count=1`
- `go build ./cmd/ntm`
- `git diff --check origin/main...HEAD`

## Scope

Source-only. No installed binary, live NTM session/config/manifests/panes, or identities changed.

### DIFF
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -3215,6 +3215,7 @@ Shell Integration:
 		// Robot-mail-check handler for Agent Mail inbox (bd-adgv)
 		if robotMailCheck {
 			opts := robot.MailCheckOptions{
+				Client:        newAgentMailClient(mailProject),
 				Project:       mailProject,
 				Agent:         mailAgent,
 				Thread:        mailThread,
--- a/internal/resilience/monitor.go
+++ b/internal/resilience/monitor.go
@@ -29,7 +29,21 @@ var (
 	checkSessionFn   = health.CheckSession
 	displayMessageFn = tmux.DisplayMessage
 	isChildAliveFn   = process.IsChildAlive
+	paneExistsFn     = paneExistsInSession
 )
+
+func paneExistsInSession(session, paneID string) (bool, error) {
+	panes, err := tmux.GetPanes(session)
+	if err != nil {
+		return false, err
+	}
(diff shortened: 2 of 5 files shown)


Dicklesworthstone:
Thanks — both defects verified on current main. --robot-mail-check really did build a bare agentmail.NewClient(), skipping the config-file URL/token and (the bigger practical loss) the hydrated per-project registration tokens every other CLI mail path gets; and the resilience monitor really did inject restart keystrokes at a remembered pane ID with no membership check, which matters because pane IDs are only unique per tmux server lifetime.

Per this repo's policy, outside PRs aren't merged directly; I read the diff as a report and reimplemented both fixes independently on main in 59569ca6, with two deliberate divergences worth noting:
- The unavailable-server response keeps ErrCodeDependencyMissing (robot error codes are a consumer contract; switching to INTERNAL_ERROR would break anyone matching on it) — the remediation text now points at the configured URL/token as well.
- A *transient* pane-membership query failure skips the restart fail-closed but keeps the binding registered for the next cycle; only a confirmed-missing pane retires it. The PR deleted the binding in both cases, which would permanently deregister an agent over one flaky tmux query.

Your injected-client test design (a bearer-token-enforcing fake server) and the stale-pane zero-keys assertion both carried over in spirit. Closing with the work landed; thanks for the sharp pair of findings.

Issue #69, opened 2026-03-01
ntm spawn does not reuse agent mail identities on session respawn


RogerNavelsaker:
Summary

When a session is killed with ntm kill <session> and recreated with ntm spawn <session> --cc=N --gmi=N, each agent receives a new randomly generated name in the Agent Mail service instead of reusing the names from the previous
session. This breaks any ongoing mail threads, contact references, or coordination context that referenced the previous agent identities.

Steps to reproduce

ntm spawn myproject --cc=1 --gmi=1
# observe: "Registered agent pane 1 as AmberRidge, pane 2 as MistyDeer"

ntm kill myproject --force
ntm spawn myproject --cc=1 --gmi=1
# observe: "Registered agent pane 1 as GreenLantern, pane 2 as CrimsonLynx"
# Expected: AmberRidge and MistyDeer reused

Root cause

ntm already persists the slot→name mapping in ~/.config/ntm/sessions/<session>/<session_lower>/agent_registry.json:

{
  "agents": {
    "myproject__cc_1": "AmberRidge",
    "myproject__gmi_1": "MistyDeer"
  },
  "pane_id_map": { "%10": "AmberRidge", "%11": "MistyDeer" }
}

However, on respawn, ntm generates fresh names rather than consulting this file. The Agent Mail server's register_agent tool accepts an explicit name parameter as its highest-priority identity anchor — if ntm passed the stored name,
the server would reuse the existing agent record (updating last_active_ts only).

Expected behaviour

When agent_registry.json exists for a session name and contains a mapping for a given slot (e.g. myproject__cc_1), ntm spawn should pass the stored name to register_agent rather than generating a new one. This makes respawn idempotent
  with respect to agent mail identity.

Suggested implementation

In the spawn path, before registering each agent pane with Agent Mail:
1. Check whether agent_registry.json exists for the session.
2. If a stored name exists for the slot being spawned, pass name=<stored_name> to register_agent.
3. If register_agent returns a conflict (name taken by a different project), fall back to generating a new name.

No new flags or config knobs should be needed — the behaviour should be automatic and opt-out via a flag (e.g. --fresh-identity) for cases where a clean slate is intentional.

Workaround (current)

Avoid ntm kill + ntm spawn cycles. Instead, use ntm interrupt <session> to stop agents and restart them within the existing tmux session — pane IDs are preserved, so Agent Mail identities survive.

PR #280, opened 2026-08-31
fix(agentmail): bind spawned identities to bare pane generation

luminous-maximus:
## Summary
- carry each spawned bare `%N` pane as `X-Tmux-Pane` on Agent Mail registration requests
- refuse composite pane addresses and stop overwriting server generation receipts with plain names
- verify the structured 0.3.31 binding receipt before publishing registry authority, and mirror its exact bytes into worktree project namespaces
- re-register reused identities so their binding follows the new pane generation

## Verification
- `go test ./internal/agentmail -count=1`
- focused spawn/registration suite: `go test ./internal/cli -run TestSpawnIdentityCoordinator\|TestSpawnPublishesIdentityBeforeLaunch\|TestAgentMail.*Identity\|Test.*AgentMail.*Register -count=1 -timeout=5m`
- `go build ./cmd/ntm`

The full `internal/cli` package also ran to completion; its failures were pre-existing tmux E2E cases whose fake agent panes remain at `zsh`/`sh` and are therefore correctly rejected as `PANE_AGENT_DEAD`, unrelated to this diff.

### DIFF
--- a/internal/agentmail/client.go
+++ b/internal/agentmail/client.go
@@ -534,6 +534,9 @@ func (c *Client) callTool(ctx context.Context, toolName string, args map[string]
 	}
 
 	req.Header.Set("Content-Type", "application/json")
+	if paneID, _ := ctx.Value(tmuxPaneContextKey{}).(string); paneID != "" {
+		req.Header.Set("X-Tmux-Pane", paneID)
+	}
 	if c.bearerToken != "" {
 		req.Header.Set("Authorization", "Bearer "+c.bearerToken)
 	}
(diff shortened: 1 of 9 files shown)


Dicklesworthstone:
Thanks — the direction here was verified against the mcp-agent-mail reference implementation and is real: ntm registration never told the server which pane an identity lives in, so the server's pane-binding features (generation receipts, pane-identity reuse, binding liveness) never engaged, and ntm's plain-name identity writes clobbered any structured receipt the server had written.

Per this repo's policy, outside PRs aren't merged directly; I read the diff as a report and reimplemented the direction independently on main in e3eeed31, diverging on three points after reading the server source:
- The pane travels as the `pane_id` *tool argument*, which register_agent/create_agent_identity accept directly (the server docs even say HTTP clients should pass it explicitly). The X-Tmux-Pane header is the server's fallback injection for clients that cannot set arguments — ntm can, so no context-key header plumbing. Non-bare addresses are omitted rather than failing the registration, since the binding is optional.
- publishIdentity preserves a matching server receipt and mirrors its exact bytes into worktree namespaces (your MirrorVerifiedIdentity idea, kept), but a *missing* receipt degrades to the existing plain-name writes instead of failing the spawn — the server's receipt write is best-effort in its own source, and older servers / split-filesystem setups would otherwise fail every pane.
- The dead-slot reuse path makes the re-registration you proposed (name + new pane id, tokens hydrated) but strictly best-effort, keeping the #69 offline-reuse contract; a refusing server still yields the recovered name, with a regression test pinning that.

The receipt-vs-plain-name distinction and the reuse-rebinding observation were both genuinely good catches. Closing with the work landed; thanks.

PR #281, opened 2026-08-31
fix(assign): respect persona roles when matching work

luminous-maximus:
## Summary
- resolve persona role tags from each pane's durable persona variant
- exclude reviewer-only panes from `--template=impl` assignment and implementer-only panes from `--template=review`
- preserve compatibility for untagged/model-only and explicitly dual-role panes
- apply the same eligible candidate set to balanced/allocation and legacy strategies, including auto-reassignment

## Verification
- `go test -short ./...`
- `go test -short ./internal/cli ./internal/assign`
- `go build ./cmd/ntm`
- `gofmt -l internal/cli/assign.go internal/cli/assign_test.go` (clean)
- `git diff --check`

Tracks agent-factory-0t6k.

### DIFF
--- a/internal/cli/assign.go
+++ b/internal/cli/assign.go
@@ -257,6 +257,16 @@ func filterAssignAgentsByTemplate(agents []assignAgentInfo, template string) []a
 	return filtered
 }
 
+func roleEligibleAssignAgents(agents []assignAgentInfo, projectDir, template string) ([]assignAgentInfo, error) {
+	withRoles, err := applyAssignPersonaRoles(agents, projectDir)
+	if err != nil {
+		return nil, err
+	}
+	return filterAssignAgentsByTemplate(withRoles, template), nil
+}
+
+var checkAssignCycles = CheckCycles
+
 func newAssignCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "assign [session]",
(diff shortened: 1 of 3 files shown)


Dicklesworthstone:
Thanks — the core observation is right: assignment treated every idle pane as interchangeable, so with the default --template=impl a reviewer-persona pane could receive implementation work, and your four integration points (planning, retry, reassignment, auto-reassignment) were exactly the idle-pool consumers that needed the gate.

Per this repo's policy, outside PRs aren't merged directly; I read the diff as a report and reimplemented the direction independently on main in 379fae55, with one deliberate divergence: the PR's predicate matched only literal "implementer"/"reviewer" persona *tags*, but ntm's builtin implementer/reviewer personas carry tags like "implementation"/"review", not the role tokens — so the filter would never have engaged for the builtins users actually spawn. The landed version resolves roles from three explicit signals: pane title tags, the persona *name* bound to the pane's variant (covering the builtins), and the persona's tags — while still ignoring broad descriptive tags so the mixed-focus architect stays dual-eligible. Same exclusion semantics as yours (reviewer-only out of impl, implementer-only out of review, untagged/model-only/dual-role eligible everywhere, other templates unfiltered), applied at all four call sites with PERSONA_ERROR fail-closed on a corrupt registry. Closing with the work landed; the eligible-set framing made this straightforward to build on.

PR #282, opened 2026-08-31
fix(assign): reconcile stale plans with live bead lifecycle

luminous-maximus:
## Problem

`ntm assign --watch` treats `bv --robot-plan` as live lifecycle authority. If a tracked plan item is closed mid-session, it disappears from both `br ready` and `br list --status open`; label enrichment then returns a fatal error and kills the watch loop. The same stale-plan shape can dispatch blocked or in-progress/review work.

## Fix

When a plan row disappears from both open-only label surfaces, resolve it once against authoritative `br list --all` lifecycle state:

- closed, blocked, in_progress, deferred, or any other non-open status: skip as satisfied/ineligible;
- still open but missing label/type coverage: keep failing closed;
- absent from the complete tracker surface: keep failing closed.

The all-lifecycle read only runs for rows missing from the two existing open surfaces, so the normal two-call path is unchanged.

## Tests

- lifecycle table covers closed/blocked/in_progress/deferred;
- live-open-but-unverified remains refused;
- built-process E2E drives real `ntm assign --auto` with a stale-open BV plan plus live-closed Beads row and asserts exit 0, success, zero assignments/failures, no prompt/mail, and unchanged state;
- missing-ID built-process E2E remains fail-closed;
- deleting the non-open skip kills the lifecycle tests;
- `go test ./internal/bv` and tagged built-process E2E pass.

### DIFF
--- a/e2e/atomic_assignment_cli_e2e_test.go
+++ b/e2e/atomic_assignment_cli_e2e_test.go
@@ -5084,15 +5084,21 @@ func TestE2EAtomicAssignmentLabelEnrichmentFailuresFailClosedBuiltProcess(t *tes
 		"#!/bin/sh",
 		`printf '%s\n' "$*" >> "$NTM_E2E_BR_TRACE"`,
 		`command=`,
+		`all=false`,
 		`for arg in "$@"; do`,
 		`  case "$arg" in ready|list) command="$arg"; break ;; esac`,
 		`done`,
+		`for arg in "$@"; do [ "$arg" = "--all" ] && all=true; done`,
+		`[ "$command" = list ] && [ "$all" = true ] && command=list_all`,
 		`case "$NTM_E2E_BR_MODE:$command" in`,
 		`  ready_failure:ready) printf 'injected br ready failure\n' >&2; exit 70 ;;`,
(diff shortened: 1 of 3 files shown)


Dicklesworthstone:
Thanks — verified end to end: a tracked plan item closed after the bv snapshot vanishes from both br ready and br list --status open, and the label-enrichment error killed assign --watch outright; treating bv --robot-plan as lifecycle authority was the real defect, and resolving only the missing rows against br list --all (skip non-open, keep failing closed for still-open or fully absent) is the right shape — conservative, and free on the normal path.

Per this repo's policy, outside PRs aren't merged directly; I read the diff as a report and reimplemented it independently on main in 8212e723: the same lifecycle reconciliation (with the plan-item filter restructured to keep planItems/planByID consistent rather than delete-and-guard), the allBeadStatusesContext reader with blank-status fail-closed, unit coverage for the closed/blocked/in_progress/deferred skip table plus both fail-closed branches, and the built-process e2e updates — the stale-closed clean-success case, the three-surface diagnostic, and the expected br invocation trace. Your mutation-kill check (deleting the non-open skip kills the lifecycle tests) holds for the landed version too. Closing with the work landed; this was a well-engineered report.

PR #284, opened 2026-08-31
fix(agent-mail): scope controller identities to canonical project

lifeofgurpreet:
## Problem

A fleet using a canonical non-filesystem Agent Mail project key could not use NTM worker/controller identity setup safely: NTM derived physical worktree keys, normal spawn paths could override the public token, and controller launch had no identity bootstrap.

## Change

- Add an invocation-scoped canonical Agent Mail project-key override without changing default behavior.
- Preserve escaped resource URI keys and scope identity files to the canonical key.
- Prefer MCP_AGENT_MAIL_TOKEN over stale configured tokens across spawn, routing, and dashboards.
- Publish controller identity before its agent command is sent.

## Verification

- Full internal agentmail suite.
- Focused CLI, robot, and dashboard tests.
- go build ./cmd/ntm.
- git diff --check.

This does not claim to authenticate direct robot actuation; that is a separate control-plane issue.

### DIFF
--- a/internal/agentmail/client.go
+++ b/internal/agentmail/client.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -171,8 +172,12 @@ func NewClient(opts ...Option) *Client {
 		},
 	}
 
-	// Check environment variables
+	// Check environment variables. MCP_AGENT_MAIL_TOKEN is the public MCP
+	// convention; AGENT_MAIL_TOKEN remains the explicit NTM override when both
+	// are present.
 	if token := os.Getenv("AGENT_MAIL_TOKEN"); token != "" {
+		c.bearerToken = token
+	} else if token := os.Getenv("MCP_AGENT_MAIL_TOKEN"); token != "" {
 		c.bearerToken = token
 	}
 	if baseURL := os.Getenv("AGENT_MAIL_URL"); baseURL != "" {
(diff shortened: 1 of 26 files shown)


Dicklesworthstone:
Thanks for the detailed report. Per this repo's policy, outside PRs aren't merged directly — I read the diff as intel and verified each strand against current main and the mcp-agent-mail source. Two contained findings were real and landed independently on main in 10dc2ba1:

- **Controller pane targeting** (your "pane indices repeat; reusing pane 1 can overwrite a running agent" observation — the sharpest catch in the PR): verified that buildControllerResponse reused the first index-1 pane from any window and typed the launch command into it unguarded. The landed fix keeps the in-window UX rather than adopting the detached-window redesign: pane 1 is reused only when it is verifiably a bare shell pane in the session's first window; anything else gets a freshly created pane.
- **Dashboard mail client semantics**: the reservation watcher forced config URL/token over the AGENT_MAIL_URL/AGENT_MAIL_TOKEN environment (opposite of every other CLI mail path) and skipped registration-token hydration; it now uses the shared newAgentMailClient constructor.

Not adopted, with reasons:
- **MCP_AGENT_MAIL_TOKEN**: checked the mcp-agent-mail source and docs — the server's convention is HTTP_BEARER_TOKEN server-side, and MCP_AGENT_MAIL_TOKEN appears nowhere in it, so this reads as a fleet-local convention rather than "the public MCP convention"; honoring an undocumented env name in the client felt like namespace risk without a contract behind it.
- **Invocation-scoped canonical project-key override**: rewriting project_key/human_key for every RPC at the client boundary via an ambient env override is action-at-a-distance that the rest of ntm's session/worktree key resolution isn't designed around; if canonical non-filesystem keys are a need, that wants an owner-level design (probably first-class config, not a client-layer rewrite). Same for the controller model/reasoning-effort flags and coordinator identity-bootstrap rework — reasonable directions, but design-scoped.

Yesterday's c79cc100 had already landed the coordinator-identity registration piece this overlaps with. Closing with the two fixes landed; the pane-1 hazard alone made this PR worth filing.

PR #286, opened 2026-09-01
fix(windows): normalize SQLite file URIs

kjgryboski:
## Summary

- serialize absolute Windows drive paths as empty-authority SQLite URIs (`file:///C:/...`) instead of treating the drive letter as a forbidden authority
- use slash-separated `path.Join` for embedded migration assets so `embed.FS` works on Windows
- cover drive, relative, UNC, non-Windows, pragma, and immediate-transaction behavior

## Root cause

NTM 1.30.0 native Windows robot commands failed before loading state with `invalid uri authority: C:%5C...`. After correcting the DSN, the same path exposed a second Windows separator bug: `embed.FS` could not open `migrations\\001_initial.sql`.

## Verification

- `go test -short ./...` — pass
- `go test ./internal/sqliteutil ./internal/state ./internal/bv` — pass
- `gofmt` / `git diff --check` — pass
- UBS on the three changed Go files — no critical findings; its lone warning was the isolated shadow scan not seeing the repository `go.mod`
- independently reviewed with no material findings
- exact v1.30.0 local source commit `da43c19cf4c87a4e6261237268ee3680e94b999d` built twice to identical SHA-256 `63e6ae5e612187cba917cedf32bf4573888f898ba19984b23569bc864aa727f7`
- that candidate returns healthy structured `--robot-status` on Windows against the existing state store

No tmux or orchestration behavior changes. WSL remains the launch-capable control plane; this patch restores native Windows state/robot inspection.


Dicklesworthstone:
Thanks — both defects verified exactly as reported on current main: url.URL rendered absolute drive paths with the drive letter in the URI authority (the driver's `invalid uri authority: C:%5C...`), and ReadMigration used filepath.Join for embed.FS lookups, which produces backslash paths embed.FS cannot open on Windows.

Per this repo's policy, outside PRs aren't merged directly; I read the diff as a report and reimplemented the fixes independently on main in e96f47b2: Windows paths normalized to forward slashes with absolute drive paths getting the /C:/... empty-authority form (non-Windows paths pass through byte-for-byte, since backslash is a legal POSIX filename byte), path.Join for the embedded migrations, and tests covering drive, drive-relative, relative, UNC, and non-Windows shapes plus DSN authority/pragma/_txlock preservation — the same case matrix as yours, which was well chosen. The GOOS-threaded helper for host-independent testing was a good idea and survived reimplementation intact. Closing with the work landed; thanks for restoring native Windows state inspection.

PR #289, opened 2026-09-02
fix(assign): skip packet tracking beads

luminous-maximus:
## Summary
- treat the packet-tracked label as a built-in operator gate
- exclude packet tracking beads from direct assign, assign --watch, and coordinator recommendation paths through the shared canonical gate
- add focused regressions for direct assignment and watch-loop draining; the coordinator's canonical-vocabulary regression covers every built-in label

## Verification
- targeted: go test -run 'TestOperatorGatedLabelsCanonicalVocabulary|TestPacketTrackedBeadIsExcludedFromDirectAssignment|TestWatchLoopShouldStopUsesFilteredActionableCandidates|TestRecommendationPassesSemanticGatesUsesCanonicalOperatorLabels' ./internal/bv ./internal/cli ./internal/coordinator
- full: go test -short ./... (receipt-bound run in progress)

Tracking: agent-factory-yv4v

### DIFF
--- a/internal/bv/bead_status_test.go
+++ b/internal/bv/bead_status_test.go
@@ -285,6 +285,7 @@ func TestOperatorGatedLabelsCanonicalVocabulary(t *testing.T) {
 		"needs-operator",
 		"operator-action",
 		"operator-gated",
+		"packet-tracked",
 	}
 	got := OperatorGatedLabels()
 	if !reflect.DeepEqual(got, want) {
--- a/internal/bv/bead_status_test.go
+++ b/internal/bv/bead_status_test.go
@@ -416,6 +416,28 @@ func TestValidateBeadAssignmentAuthorizationAllowsExplicitStaleOwnershipState(t 
 	}
 }
 
+func TestValidateBeadAssignmentAuthorizationRejectsPacketTrackedBead(t *testing.T) {
+	t.Parallel()
+
+	details := &BeadAssignmentDetails{
+		ID:     "agent-factory-packet",
+		Status: "open",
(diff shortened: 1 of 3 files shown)


Dicklesworthstone:
Thanks for the report and for the careful diff. This project doesn't merge outside PRs, so I'm closing this one, but I read it end to end.

I can't land the change itself: `packet-tracked` is not a label convention that ntm (or br) defines anywhere, so hard-coding it into the built-in operator-gate vocabulary would bake one external workflow's marker into every project's assignment policy. The built-in list is meant for the gates ntm itself documents; project-specific vocabulary is exactly what `[assign] operator_gated_labels` exists for.

What you're after already works on every automated dispatch path without a code change. In the project's ntm config:

```toml
[assign]
operator_gated_labels = ["packet-tracked"]
```

That set is merged with the built-ins (it can only extend them, never remove one), installed process-wide via `bv.ConfigureOperatorGatedLabels` and per-project via `bv.ConfigureProjectOperatorGatedLabels`, and consulted by `ntm assign`, `ntm assign --watch`, `--robot-assign` / bulk assign, spawn-time assignment, the coordinator's semantic gate, and the final claim transaction (`ValidateBeadAssignmentAuthorizationWithOperatorGatedLabels`). Matching is case-insensitive. A bead carrying that label is reported as `operator_gated` in the skipped list on every scan, and a queue that contains only gated beads drains the watch loop, which is the behaviour your tests describe.

If there's a case where a label configured this way still gets dispatched, that's a real bug and I'd like to hear about it with the config and the command you ran.

PR #290, opened 2026-09-02
fix(bv): retry Beads write-lock timeouts

luminous-maximus:
## Summary

- Treat Beads Timed out ... waiting for write lock diagnostics as transient contention.
- Reuse the existing bounded, context-cancellable retry path.
- Keep stale in_progress plan items nonfatal across three real watch-loop scans.
- Bind the write-lock regression to three retries within each individual scan.

## Verification

- go build ./cmd/ntm
- go test ./internal/cli -run TestWatchLoopSurvivesThreeScans -count=1
- go test ./internal/bv -run 'TestRunBdContextRetriesBeadsWriteLockTimeout|TestGetActionableRecommendationsSkipsPlanRowsClosedSinceSnapshot' -count=1
- Mutation proof: removing the write-lock classifier makes TestWatchLoopSurvivesThreeScansWithTransientBeadsWriteLockTimeout fail with context deadline exceeded.

The full short internal/bv + internal/cli run has two failures already present at upstream main 75a6f92c: TestGetReadySnapshotContextUsesOneDirectReadyResult and TestGetBlockedSnapshotContextRetainsRestrictiveEvidence. Both reject the newly added --lock-timeout 5000 argument; this branch adds no failure delta.

### DIFF
--- a/internal/bv/bv.go
+++ b/internal/bv/bv.go
@@ -2294,7 +2294,9 @@ func isNoBeadsDBError(streams ...string) bool {
 
 func isTransientBeadsDBError(streams ...string) bool {
 	s := strings.ToLower(strings.Join(streams, "\n"))
-	return strings.Contains(s, "database is busy") || strings.Contains(s, "database is locked")
+	return strings.Contains(s, "database is busy") ||
+		strings.Contains(s, "database is locked") ||
+		strings.Contains(s, "waiting for write lock")
 }
 
 func transientBeadsDBBackoff(attempt int) time.Duration {
--- a/internal/bv/bv_helpers_test.go
+++ b/internal/bv/bv_helpers_test.go
@@ -322,6 +322,43 @@ func TestRunBdContextCancelsDuringTransientRetryBackoff(t *testing.T) {
 	}
 }
 
+func TestRunBdContextRetriesBeadsWriteLockTimeout(t *testing.T) {
+	dir := t.TempDir()
+	if err := os.Mkdir(filepath.Join(dir, ".git"), 0o755); err != nil {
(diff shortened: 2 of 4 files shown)


Dicklesworthstone:
Thanks for this — the diagnosis is right, and the reproduction with br's exact `CONFIG_ERROR: Timed out after 5000ms waiting for write lock ...` text was useful. This project doesn't merge outside PRs, so I'm closing this one, but the fix is in: 8571b2cb (ships in the next release).

What landed, and where it goes further than the one-line classifier change:

- br's write-lock timeout is now classified as workspace contention using the same predicate br applies internally (`Timed out after` + `waiting for write lock`). Other `CONFIG_ERROR`s — including a genuinely unwritable lock file — remain terminal and are never retried, which a matching test pins.
- The retry ladder is now bounded exponential backoff with uniform jitter (`[ceiling/2, ceiling]`, 50ms doubling to an 800ms cap) so several ntm processes contending on the same workspace stop retrying in lockstep, and the whole sequence is bounded by a 20s total deadline on top of the per-attempt cap and the caller's context.
- The bv command runner shares the same classifier and budget logic instead of its own copy of the SQLite string match.
- The two `work_readiness` tests you noticed failing on `main` were fixed in the same commit (their fake br didn't accept the `--lock-timeout 5000` argv).

Regression coverage: `TestRunBdContextRetriesBeadsWriteLockTimeoutThenSucceeds`, `TestRunBdContextDoesNotRetryTerminalConfigError`, `TestRunBdContextStopsRetryingContentionAtTotalDeadline`, plus classifier and backoff-bounds tests in `internal/bv/bv_helpers_test.go`.

PR #291, opened 2026-09-02
fix(spawn): refuse occupied launch panes

luminous-maximus:
## Problem

`ntm spawn` reused an existing session pane without verifying that the pane still ran a shell. `tmux send-keys` then succeeded even when an interactive agent CLI was listening, submitting the raw launch command as a prompt.

Reproduction on v1.31.0:

1. Start an exact disposable pane with `tmux new-session -d -s wsnk-repro cat`.
2. Run `ntm spawn wsnk-repro --cc=1 --no-user --no-hooks --no-cass-context --no-recovery --privacy` against that existing session.
3. The command returns success, `pane_current_command` remains `cat`, and capture-pane contains the full Claude launch line (twice) rather than a launched process.

The captured line included `--dangerously-skip-permissions`; it was harmless only because `cat` received it as input.

## Fix

- Preflight every assigned pane for every agent family before launching any member of the batch.
- Re-read the exact physical CLI pane after slow preparation and immediately before `send-keys`.
- Refuse on an occupied, missing, or unreadable target. Never retarget.
- Apply the same all-agent baseline rule to robot spawn and its launch helper.

With this branch, the same live probe returns rc=1 naming pane `%1088` and current command `cat`; the pane capture remains empty.

## Tests

- `go test -short ./internal/cli ./internal/robot ./internal/tmux -count=1` — PASS.
- `go vet ./...` — PASS.
- `go test -short ./...` — all packages pass except two `internal/bv` tests already failing unchanged on pristine base 75a6f92c: `TestGetReadySnapshotContextUsesOneDirectReadyResult` and `TestGetBlockedSnapshotContextRetainsRestrictiveEvidence` (their fake br scripts do not accept the new `--lock-timeout 5000` argv).

Workshop tracking: agent-factory-wsnk.

### DIFF
--- a/internal/cli/spawn.go
+++ b/internal/cli/spawn.go
@@ -909,24 +909,60 @@ func validateSpawnPaneCapacity(panes []tmux.Pane, startIdx, agentCount int) erro
 	return nil
 }
 
-func validateSpawnGrokPaneBaselines(panes []tmux.Pane, startIdx int, agents []FlatAgent) error {
+func validateSpawnPaneBaselines(panes []tmux.Pane, startIdx int, agents []FlatAgent) error {
 	for agentOffset, launch := range agents {
-		if launch.Type != AgentTypeGrok {
-			continue
-		}
 		paneOffset := startIdx + agentOffset
 		if paneOffset < 0 || paneOffset >= len(panes) {
 			return fmt.Errorf(
-				"no assigned pane for Grok Build agent %d at offset %d",
+				"no assigned pane for %s agent %d at offset %d",
+				launch.Type,
 				launch.Index,
 				paneOffset,
 			)
 		}
 		if err := tmux.ValidatePaneLaunchBaseline(panes[paneOffset]); err != nil {
-			return fmt.Errorf("validate Grok Build agent %d: %w", launch.Index, err)
+			return fmt.Errorf("validate %s agent %d: %w", launch.Type, launch.Index, err)
 		}
 	}
 	return nil
+}
+
+type spawnPaneBaselineLister func(context.Context, string) ([]tmux.Pane, error)
+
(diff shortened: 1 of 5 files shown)


Dicklesworthstone:
Thanks — this was a real hole, and the `cat` reproduction made it easy to confirm: only Grok launches were checking that the target pane was still an idle shell, so every other agent type could have its launch line (including `--dangerously-skip-permissions`) typed into a live process while spawn reported success. This project doesn't merge outside PRs, so I'm closing this one, but the fix is in: e4718530 (ships in the next release).

What landed:

- `tmux.ValidatePaneLaunchBaseline` keeps deciding from tmux's own `pane_current_command` (the foreground process on the pane's tty — a fresh probe confirms it flips from the shell to the foreground child and back deterministically), now wraps a `tmux.ErrPaneOccupied` sentinel, and explains how to free the pane.
- New `tmux.ValidatePaneLaunchBaselineLiveContext` re-reads the exact pane by ID immediately before launch and fails closed when the pane is occupied, missing, or unreadable. Launches are never retargeted.
- `ntm spawn` preflights every pane of the batch for every agent type before launching any of them (no partial launches), then re-validates the exact pane after identity preparation and right before `send-keys`. The error names the agent, pane and running command and points at `ntm add` for growing a session that already has running panes.
- Robot spawn's existing-pane and final-topology preflights now cover every agent family, `launchAgent` validates the snapshot plus a live re-read for every agent type, and the hints are agent-generic.

Regression coverage in `internal/tmux`, `internal/cli` and `internal/robot` (occupied Claude/Codex targets, all five agent types through `launchAgent`, robot spawn refusing an occupied existing pane before any split/layout/launch, and the live re-read's refuse-don't-retarget cases).

PR #293, opened 2026-09-02
fix(assign): skip packet tracking beads

luminous-maximus:
# Proposed upstream pull request

Repository: `Dicklesworthstone/ntm`

Title: `fix(assign): skip packet tracking beads`

## Summary

- Add `packet-tracked` to ntm's built-in operator-gated label vocabulary.
- Exclude a packet-tracked bead from direct assignment, auto-assignment,
  coordinator scans, and `assign --watch` through the shared assignment gate.
- Preserve the final guarded claim as a second fail-closed check if a stale
  recommendation reaches execution.

## Problem

Mission packet `agent-factory-el73` was filed as a tracking bead even though a
mission packet is worked as a whole in its own window. The assignment rail saw
an open, unblocked bead and offered it to an iris worker. A comment convention
cannot stop that recurrence because ntm plans from authoritative Beads fields.

George's 2026-09-02 ruling removes tracking beads from the mission-packet v2
contract. `packet-tracked` remains a defense for an already-created bead or any
other non-rail work represented in the Beads store.

## Implementation

ntm already has one canonical operator-gate vocabulary and routes assignment
planning, watch completion, newly unblocked work, and the final transactional
claim through it. Adding `packet-tracked` there gives every rail path the same
answer without a packet-specific branch in the coordinator.

The regressions prove that:

- three consecutive scans produce no ready candidate for an open
  `packet-tracked` bead;
- watch mode treats a queue containing only that bead as drained;
- the final authorization gate returns the typed assignment-ineligible error;
- the built-in label cannot be removed by project configuration.

## Verification

- `go test -short ./internal/bv ./internal/cli -count=1` with an isolated,
  empty user-persona config
- the Agent Factory carrier's mission-packet contract probe constructs a valid
  v2 packet, adds `references.bead`, and proves the live validator refuses it.

Upstream base: `29be5f2b542ce634c932c4b47d8b8a3c84f70c2a`

Patch head: `7af568ff8f57083a6ca1899a369ba657f439949b`

### DIFF
--- a/internal/bv/bead_status_test.go
+++ b/internal/bv/bead_status_test.go
@@ -285,6 +285,7 @@ func TestOperatorGatedLabelsCanonicalVocabulary(t *testing.T) {
 		"needs-operator",
 		"operator-action",
 		"operator-gated",
+		"packet-tracked",
 	}
 	got := OperatorGatedLabels()
 	if !reflect.DeepEqual(got, want) {
--- a/internal/bv/bead_status_test.go
+++ b/internal/bv/bead_status_test.go
@@ -416,6 +416,28 @@ func TestValidateBeadAssignmentAuthorizationAllowsExplicitStaleOwnershipState(t 
 	}
 }
 
+func TestValidateBeadAssignmentAuthorizationRejectsPacketTrackedBead(t *testing.T) {
+	t.Parallel()
+
+	details := &BeadAssignmentDetails{
+		ID:     "agent-factory-packet",
+		Status: "open",
+		Labels: []string{"packet-tracked"},
+	}
+	err := ValidateBeadAssignmentAuthorizationWithOperatorGatedLabels(
+		details,
+		BeadAssignmentAuthorization{BeadID: details.ID},
+		nil,
+	)
+	var eligibilityErr *AssignmentEligibilityError
+	if !errors.Is(err, ErrBeadAssignmentIneligible) || !errors.As(err, &eligibilityErr) {
+		t.Fatalf("authorization error=%v, want typed assignment-ineligible error", err)
+	}
+	if !reflect.DeepEqual(eligibilityErr.OperatorLabels, []string{"packet-tracked"}) {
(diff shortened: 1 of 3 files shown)


Dicklesworthstone:
Closing this one — it is the same change as #293's predecessor, PR #289, which I closed on 2026-09-02 with the reasoning below, and nothing in the new description changes it. This project doesn't merge outside PRs in any case, but I read the diff end to end again.

The position is unchanged: `packet-tracked` is not a label convention that ntm or br defines anywhere, so hard-coding it into the *built-in* operator-gate vocabulary would bake one external workflow's marker into every project's assignment policy. The built-in list is for the gates ntm itself documents; project-specific vocabulary is exactly what `[assign] operator_gated_labels` exists for, and I re-verified today that it is still wired the way it needs to be.

In the project's ntm config:

```toml
[assign]
operator_gated_labels = ["packet-tracked"]
```

That set is merged with the built-ins — `config.mergeOperatorGatedLabels` can only extend them, never remove one — installed process-wide via `bv.ConfigureOperatorGatedLabels` and per-project via `bv.ConfigureProjectOperatorGatedLabels`, and consulted by `ntm assign`, `ntm assign --watch`, `--robot-assign` / bulk assign, spawn-time assignment, the coordinator's semantic gate, and the final claim transaction. Matching is case-insensitive. A bead carrying the label is reported as `operator_gated` in the skipped list on every scan, and a queue containing only gated beads drains the watch loop — which is precisely the behaviour your three regressions describe.

The one part of the new description that would change my answer is the sentence about George's 2026-09-02 ruling removing tracking beads from the mission-packet v2 contract. That is a reason for the *packet* contract to stop producing these beads, which it now does, and a reason for your carrier to configure the gate — it is not a reason for ntm's built-in vocabulary to grow an entry for a marker ntm has no other knowledge of.

If you configure it as above and a labelled bead still gets dispatched on any rail, that is a real bug and I want it: please open an issue with the config file, the exact command, and the scan output showing the bead in `recommendations` rather than `skipped`.

Issue #263, opened 2026-08-23
e2e: robot assign/bulk-assign suites fail on hosts where seeded beads are invisible to br ready/br list (environmental, pre-existing)


Dicklesworthstone:
While gating the v1.30.0 release, the primary/bulk e2e suites failed on one Linux build host with a consistent signature:

```
read actionable Beads work: actionable bead labels could not be verified:
open actionable plan item "rae2e-vno" was absent from both br ready and br list --status open
```

Affected families: `TestE2ERobot{Assign,Spawn}TerminalContractsBuiltBinary`, `TestE2ERobotCapabilitiesBuiltBinaryBudgets`, `TestE2ERobotReplay*`, `TestE2ERobotInterrupt*`, `TestE2ESpawnAssignmentPartialCoverageBuiltBinary`, `TestE2E_RobotFormatSelection`, `TestE2E_RobotBulkAssign_*` (incl. `WithFromBV`).

Evidence this is environmental, not a v1.30.0 regression:

- The identical subset fails the same way at **v1.29.3** on the same host (fresh clone, same command).
- Reproduced with both br 0.3.2 and br 0.4.0 on PATH.
- The same suites pass on other machines (macOS dev box; and the mock-mode subsets that don't shell out to a real `br`).

Two more environment-sensitive tests failed at BOTH tags on the same host (headless, no tty):

- `internal/tui/theme TestDetectDarkBackgroundUsesColorFGBG`
- `internal/checkpoint TestRunGitCommandReturnsWhenDescendantHoldsStdout` (race mode)

Also noteworthy: a stale, schema-0 4KB `beads.db` scaffold (left behind by an interrupted `br` invocation during `git checkout`) made `internal/bv TestGetTriage` fail with "IssueCount should not be 0" until the scaffold was moved aside — bv prefers an (empty) DB over the JSONL. The e2e failures above persist independently of that fix, so they are a separate mechanism, likely in how the seeded temp-project's beads are created/read by the real `br`/`bv` on this host (global br daemon/config interplay is the prime suspect).

Worth root-causing so the release gate can go fully green on a stock Linux builder: either make the e2e seeding hermetic against global br state, or detect-and-report the environmental precondition instead of failing the assertion path.

Dicklesworthstone:
Investigation status (no fix landed — findings below are verified against source, but the mechanism can only be confirmed on the failing host).

**Verified in ntm at HEAD:**
[...]

PR #294, opened 2026-09-02
fix: retry safe composer dispatch failures

luminous-maximus:
## Summary
- recognize deployed Codex 0.152.x and Claude Code 2.1.x composer frames with contract fixtures
- distinguish pre-message composer/readiness failures from genuinely ambiguous post-send failures
- return safe refusals to pending and retry them with bounded exponential backoff
- preserve the existing fail-closed, at-most-once behavior for true dispatch ambiguity

## Tests
- targeted tmux, dispatch, robot, and coordinator regression tests pass
- broad package run has two failures that reproduce unchanged at pinned upstream main: `TestGetSendIdempotency_PreflightFailureReleasesClaim` and live-spawn `TestSpawnOptions_MultipleAgentTypes`

Tracking: agent-factory-qt1k

### DIFF
--- a/internal/cli/assign.go
+++ b/internal/cli/assign.go
@@ -3066,16 +3066,26 @@ type cliAtomicPaneDispatchPort struct {
 	redactionConfig redaction.Config
 	observer        assignSessionObserver
 	bypassIdleGate  bool
+	listPanes       func(context.Context, string) ([]tmux.Pane, error)
+	deliverer       dispatchsvc.Deliverer
 }
 
 func (p *cliAtomicPaneDispatchPort) prepare(ctx context.Context, req assignment.DispatchRequest) (*dispatchsvc.Service, *dispatchsvc.Prepared, error) {
-	panes, err := tmux.GetPanesContext(ctx, p.session)
+	listPanes := p.listPanes
+	if listPanes == nil {
+		listPanes = tmux.GetPanesContext
+	}
+	panes, err := listPanes(ctx, p.session)
 	if err != nil {
 		return nil, nil, fmt.Errorf("load dispatch topology: %w", err)
 	}
+	deliverer := p.deliverer
+	if deliverer == nil {
+		deliverer = dispatchsvc.TMUXDeliverer{}
+	}
 	service, err := dispatchsvc.NewService(dispatchsvc.Ports{
 		Redactor: shellFinalMessageRedactor(p.redactionConfig), Protocols: shellDispatchProtocolPlanner{},
-		Deliverer: dispatchsvc.TMUXDeliverer{},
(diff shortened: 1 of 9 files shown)


Dicklesworthstone:
Closing — this project doesn't merge outside PRs — but the core of this is right and it is implemented in 8563fc5d. Thanks for the careful separation of the two failure classes; that distinction is the whole value here.

**What I verified.** `TMUXDeliverer.Deliver` does its context check, protocol validation, dead-agent-pane refusal, `ComposerReadyForDelivery` gate and `ClearComposerForDelivery` all *before* the `switch delivery.Protocol` that types the first byte. So every refusal raised in that prologue is provably pre-actuation. `internal/assignment` already had the `GuaranteeNoActuation` / `IsGuaranteedNoActuation` marker and the coordinator's Agent Mail port already used it for typed tool rejections — the gap was purely that the tmux delivery path had no way to say "nothing was sent", so a composer or readiness refusal fell into the outcome-unknown bucket and permanently failed the assignment.

**What landed.** `internal/dispatch` grows `GuaranteeNoDeliveryActuation` / `IsGuaranteedNoDeliveryActuation`, the prologue refusals carry that mark, and `cliAtomicPaneDispatchPort.Dispatch` translates it into `assignment.GuaranteeNoActuation`. Anything at or after the protocol switch is untouched and still fails closed — the at-most-once barrier is unchanged, which was the constraint I cared most about preserving. Your `coordinatorDispatchRetryReady` backoff landed too, near enough as written: 5s initial, doubling, capped at a minute, anchored on the durable `DispatchStartedAt`/`DispatchAttempts` so it survives a restart, and skipped entirely for a row that has not recorded a dispatch error yet. Retryable should not mean retry-every-cycle-forever.

**What I did not take.** The `listPanes`/`deliverer` injection points on `cliAtomicPaneDispatchPort` — they exist only to make the port testable, and I would rather test the classification where it is decided (the deliverer prologue) than widen a production struct for it. The composer contract fixtures for Codex 0.152.x and Claude Code 2.1.x are not in either; the composer work from #300 landed separately in ed68967c / d63151fb and the fixtures would need re-deriving against that.

Tests assert the marker survives wrapping, that an unmarked error stays outcome-unknown, that a dead-agent-pane refusal and a protocol rejection both come back marked, and the backoff's boundaries including the no-error and nil cases.

On the two failures you flagged as reproducing at pinned upstream main: `TestSpawnOptions_MultipleAgentTypes` still fails here for the same environmental reason and is tracked under #263; `TestGetSendIdempotency_PreflightFailureReleasesClaim` passes in isolation and only fails under a loaded parallel run.

Ships in the next release.

PR #296, opened 2026-09-03
fix(coordinator): reap stranded dispatch rows

luminous-maximus:
Summary

- Terminalize a non-sent assignment when its physical tmux pane is absent from the coordinator fresh-session snapshot; the persisted fail_reason names the missing target.
- Bound known pre-actuation dispatch retries at three while preserving the fail-closed ambiguity barrier for a live target whose send outcome is unknown.
- Continue the assignment pass after terminalizing a stranded row, so ready bead B can be placed in the same cycle.
- Exercise the existing operator escape hatch and prove force-clear keeps equal-generation primary and backup snapshots byte-identical.

Regression evidence

The coordinator regression is built from the preserved agent-factory incident row for agent-factory-6ndm, including dispatch_state sending, dispatch_target and occupancy_key %993, dispatch_attempts 1, generation 4775, and the original timestamps and hashes.

Tests

- PASS: focused coordinator regressions, including dead A does not block ready B and bounded known-safe retries.
- PASS: focused force-clear strict-reload and primary/backup identity test.
- PASS: full internal/coordinator package.
- PASS: go build ./cmd/ntm.
- FULL SUITE RUN: nice -n 15 make test-all returned exit 2. The five observed failures all reproduce on an untouched e4718530 upstream-main worktree:
  - internal/checkpoint TestCapturer_captureSessionState_UsesSessionSelectedPaneAcrossWindows
  - internal/cli TestRunReassignment_ToPane_Success
  - internal/cli TestRunRetryAssignments_PreservesPreviousFailReasonAndMetadata
  - internal/cli TestSessionTemplateSpawn_Builtin
  - internal/cli TestSessionTemplateSpawn_CustomUserTemplate
  The four CLI cases are live-spawn failures where fake agent panes exit to bare shells or are not prompt-ready; the checkpoint assertion is identical on the base commit.

Carrier: agent-factory-m0ip.

### DIFF
--- a/internal/assignment/store.go
+++ b/internal/assignment/store.go
@@ -28,7 +28,7 @@ const (
 	// assignmentsDirName is the directory name for assignment storage
 	assignmentsDirName     = "assignments"
 	fileExtension          = ".json"
-	assignmentStoreVersion = 9
+	assignmentStoreVersion = 10
 
 	// assignmentStoreGenerationVersion is the first schema version whose
 	// snapshots carry a monotonic persistence generation. The generation makes
@@ -135,6 +135,20 @@ type Assignment struct {
 	CompletionLeaseExpiresAt *time.Time           `json:"completion_lease_expires_at,omitempty"`
 }
 
+// DispatchFailureBlock is a durable, bead-scoped circuit breaker for a
+// server-declared permanent dispatch failure. It deliberately lives outside
+// an assignment row: operators may clear/reap that row to release its claim,
+// but doing so must not reset the dispatch-attempt ceiling and immediately
+// place the same bead again under the same failure condition.
+type DispatchFailureBlock struct {
+	BeadID       string    `json:"bead_id"`
+	FailureClass string    `json:"failure_class"`
+	Reason       string    `json:"reason"`
+	Target       string    `json:"target,omitempty"`
+	Generation   uint64    `json:"generation"`
+	BlockedAt    time.Time `json:"blocked_at"`
+}
+
 // AssignmentUpdate describes mutable assignment metadata that can be updated
 // after the initial assignment record is created.
 type AssignmentUpdate struct {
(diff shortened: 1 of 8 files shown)


luminous-maximus:
Round 4 update for agent-factory-m0ip (OPENAI implementation; ANTHROPIC re-review requested).

Head: e6bd6652bdaf045e3363e9f14e06dbade5acf82d
Tree: 009528ddd31d1f11b4e7064466e6abe1361b91c6
Parent: 2f3cf1976141a950a79636bd18d6fd04a806e07a

The durable permanent-dispatch breaker and its clear generation are now structurally keyed by `(bead, failure class)`, rather than bead-only maps carrying a class payload. The regression persists two simultaneous failure classes for one bead, proves exact-class lookup/no cross-class leakage, and verifies the explicit bead recovery override clears all active classes. The existing multi-cycle regression still proves ceiling -> terminalize -> exact-actor claim release -> row clear -> same open bead is not dispatched again, with the total transport count asserted.

Focused exact-tree tests pass x3 and `go build ./cmd/ntm` passes. Like-for-like niced full suites at this head and base e47185301a8d0149e7fac0a21c3b8b2ab9c58984 each produced 109 passing / 5 failing packages. Every apparent pin-only test was isolated at -count=3 on both trees; no stable pin-only failure exists. The stable E2E failure was 3/3 on both trees and was identically caused by the live-estate `critical_pressure` decision.

PR #294 remains open, so this branch intentionally has not rebased onto it. #294 should land first; then #296 should rebase while keeping the hard attempt ceiling before #294's retry backoff.

Dicklesworthstone:
Closing — this project doesn't merge outside PRs — but I want to be straight about the outcome: **the defect is real and confirmed, and I have not fixed it yet.** This is the one item from this batch I am leaving open rather than shipping, and it needs a decision from me about where the at-most-once barrier sits, not just an implementation.

**What I confirmed.** In `recoverPendingAssignments`, a row whose `DispatchState` is `DispatchSending` is emitted with `ErrDispatchOutcomeUnknown` and `continue`d, every cycle, forever — nothing terminalizes it. Its `OccupancyKey` keeps the pane owned, so a ready bead B targeting that pane can never be placed. That is a permanent stall, exactly as your agent-factory-6ndm row shows, and it is not self-healing. Your reading of the code is correct.

**Why I am not shipping your fix as-is.** Terminalizing a `sending` row means declaring an unknown-outcome dispatch to be a non-delivery. Your gate for that — the physical tmux pane is absent from the fresh session snapshot — is a genuinely good argument, and I think it is probably the right one: if the pane is gone, no agent can act on a message that may or may not have landed there. But "absent from the fresh snapshot" in the current code means absent from the *eligible candidate* set, which is idle-and-healthy panes, not live panes; a busy pane is missing from it too. Getting that predicate wrong turns a permanent stall into a silent double-dispatch, which is a strictly worse failure. The right shape is a live-topology check against `c.agents`, and I want to write that deliberately rather than adapt it.

The rest of the PR compounds the review: it also reworks `internal/assignment/store.go` with new persisted fields, changes force-clear snapshot handling, and bounds pre-actuation retries — four changes to the most safety-critical ledger in the system in one diff. The retry-bounding part is already in from a different angle: pre-actuation dispatch refusals are now marked and retried with bounded backoff (8563fc5d, via #294).

Two smaller things from the same neighbourhood did land in 8563fc5d, and they make this stall easier to live with in the meantime: occupied-target refusals now name the owning bead's status, age and assignment time instead of just its ID, so a stranded row is visible without diffing the ledger by hand; and reservation candidates are filtered to real repository paths, so the junk leases that accumulate around these rows stop being created.

I have opened #304 to track the stranded-row reaping itself with the analysis above, so it does not get lost behind this closed PR. The operator escape hatch remains `ntm assign --clear <bead> --force`.

On the full-suite results: the five failures you list all reproduce here on an untouched tree and are the environmental class tracked in #263 — thank you for checking them against the base commit rather than reporting them as regressions.

Dicklesworthstone:
Follow-up: this is now fixed on `main` in c2edc965 (plus 34b5d57e), and #304 is closed. The write-up is on that issue; the short version of where it landed relative to your patch:

Terminalization is gated on the live physical pane set — captured from every pane in the session observation, before the agent-type filter that drops user/unknown panes, and invalidated whenever a whole-session observation fails — rather than on the eligible-candidate set. Two things beyond the pane check turned out to be load-bearing: dispatch is an Agent Mail message addressed to an agent *name*, not to a pane, so a dead pane alone does not make the assignment unreachable. The recipient must not be live on another pane, and a mailbox probe must complete and find no message for that exact generation, before a non-delivery is asserted. When the probe *finds* the message, the row is repaired forward to `sent` against the discovered receipt instead — the landed-dispatch race.

Your analysis is what made the issue actionable, and the "evaluate against live topology, not candidates" point in particular was the right call. Thanks for the careful writeup and the preserved incident row.

PR #301, opened 2026-09-03
fix(activity): persist pane state age across invocations

luminous-maximus:
## Summary

- persist `ntm activity` state transitions in the runtime watermark store so `state_since` remains stable and `duration` grows across one-shot invocations
- classify transient pane state from the actual visible viewport instead of extra `capture-pane -S` scrollback, preventing stale Codex spinners from masking an idle composer
- carry pane width/height into the activity classifier and isolate state scopes by session, pane ID, and pane PID

## Verification

- `go test -short ./...`
- `go test ./internal/robot -run "TestObserveActivityStateSurvivesOneShotProcesses|TestStateClassifierLimitsCodexThinkingToVisiblePane" -count=1`
- `go test ./internal/status -run "TestSessionObserverClassifiesOnlyVisibleCodexViewport|TestDetermineStateCodexIdleBeatsWindowVelocity" -count=1`
- `go test ./internal/cli -run "TestShouldInitializeRobotPersistenceForActivity|TestActivityStateScopeSeparatesPaneLifetimes|TestActivity_TwoWindowPaneIndexCollision" -count=1`
- `golangci-lint run --new-from-rev=origin/main --timeout=5m` (0 issues)
- controlled 6-row tmux pane: two `ntm activity --json` calls 37s apart stayed WAITING, kept the same `state_since`, and advanced duration to 37s
- controlled `ntm assign --watch --dry-run`: initial pass and subsequent scans each found the idle pane and produced a recommendation

## Existing upstream lint debt

Full `golangci-lint run --timeout=5m` reports 25 findings, all outside the files changed by this PR. The changed-lines run is clean.

### DIFF
--- a/internal/cli/activity.go
+++ b/internal/cli/activity.go
@@ -105,9 +105,17 @@ type activityOptions struct {
 	filterPane        string
 	watchMode         bool
 	interval          time.Duration
+	watermarkStore    robot.WatermarkStore
 }
 
 func runActivity(session string, opts activityOptions) error {
+	// PersistentPreRun opens the runtime store for the real command path.
+	// Carry it explicitly into collection so direct unit-test/library calls do
+	// not depend on mutable package-global store lifecycle.
+	if opts.watermarkStore == nil {
+		opts.watermarkStore = robotStateStore
+	}
+
 	// bd-ixy2t: emit a JSON envelope on early-fail when --json is set so
 	// automation pipelines (`ntm activity --json | jq ...`) always see a
 	// parseable failure on stdout instead of a stderr "Error:" line.
@@ -263,7 +271,7 @@ func collectActivityData(session string, opts activityOptions) (*activityResult,
 
 	result := &activityResult{
 		Session:    session,
-		CapturedAt: time.Now(),
+		CapturedAt: time.Now().UTC(),
 		Agents:     make([]agentInfo, 0),
 		Summary:    make(map[string]int),
 	}
@@ -283,7 +291,9 @@ func collectActivityData(session string, opts activityOptions) (*activityResult,
 
 		// Create classifier and get state
 		classifier := robot.NewStateClassifier(pane.ID, &robot.ClassifierConfig{
-			AgentType: agentType,
+			AgentType:  agentType,
+			PaneWidth:  pane.Width,
+			PaneHeight: pane.Height,
 		})
 
 		activity, err := classifier.Classify()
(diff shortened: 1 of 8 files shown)


Dicklesworthstone:
Closing — this project doesn't merge outside PRs — but this is a good catch and all three parts are implemented, in 18ab9341 and 1f4966bc.

**Verified.** `collectActivityData` constructed a fresh `robot.NewStateClassifier` per pane per invocation, so `activity.StateSince` was necessarily "now" and `duration` was always ~0 across one-shot calls. `shouldInitializeRobotPersistence` only returns true for `--robot-*` invocations and `--schema`, so `activity` never opened the runtime store either — there was nothing to remember with. Both exactly as described.

**What landed.**

`robot.ObserveActivityState(store, scope, state, observedAt)` reconciles the classified state against a durable watermark under a new `activity_state` watermark type: unchanged state keeps its original entry time, a changed state moves it, and `CreatedAt` is preserved across transitions. Your scope key is right and is what shipped — session, pane ID, and pane PID, so a recycled tmux pane ID cannot inherit the previous process's age. A nil store degrades to the observation time rather than failing, and a store error leaves the caller on its non-durable classifier result.

`activity` now initializes persistence but skips the bv/br projection refresh — that refresh is a 10s-budget round trip and a fast pane snapshot should not pay for it. Duration is measured against the frame's own `CapturedAt` (now UTC), so every row in a frame shares one clock and a future or corrupt watermark cannot produce a negative age; I added an explicit regression for the skewed-watermark case.

The viewport half landed too, in both places: `ClassifierConfig.PaneHeight` bounds the live window to what is actually on screen, and `SessionObserver.buildPaneObservation` classifies `util.GetLastNLines(output, pane.Height)` while keeping the full capture for content-change history and diagnostics. `capture-pane -S` returns scrollback above the visible screen, so on a short pane a codex "Working" bullet that scrolled away hours ago sat inside the "live tail" indefinitely — the argument is sound and the fix is gated on tmux actually reporting a height, so it can only ever narrow, never invent.

Tests cover the stable/changed/stable watermark sequence, scope separation across pane PIDs, the nil-store and future-watermark degradations, the off-screen-vs-on-screen Codex spinner pair, the scope key's distinctness across session/pane/PID, and that `activity` initializes persistence without the projection refresh.

On the lint note: the 25 pre-existing findings outside your changed files are still there and are not from this work.

Ships in the next release.

PR #302, opened 2026-09-03
fix(coordinator): unblock daemon placement

luminous-maximus:
## Summary

- add `ntm coordinator run --reserve-files=false` and bypass the reservation port when disabled
- filter coordinator reservation candidates to real repository paths, excluding slash-bearing prose such as `13714/2m`
- make occupied-pane diagnostics report the owning bead, status, age, and assignment time
- add terminal assignment cleanup with `ntm assign --retain-claim`, plus exact malformed-project reservation release by durable identity

## Verification

- `go test -short ./...`
- `go build ./cmd/ntm`
- `go vet ./...`
- `gofmt` and `git diff --check`

Agent Factory bead: agent-factory-yqu9

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,6 +310,8 @@ ntm coordinator conflicts payments
 ntm coordinator enable auto-assign
 ntm coordinator enable digest --interval=30m
 ntm coordinator disable conflict-negotiate
+ntm coordinator run payments --reserve-files=false
+ntm assign payments --clear-pane=3 --force --retain-claim
 ```
 
 `ntm locks force-release` is approval-gated by default (`automation.force_release:
@@ -323,6 +325,12 @@ action honor the same policy setting.
 `coordinator enable` and `disable` persist the selected `--config` file, or the
 global config by default, without replacing unrelated settings or comments.
 Restart an already running `ntm coordinator run` daemon to apply a toggle.
+Pass `--reserve-files=false` to that daemon when Agent Mail reservations are
+unavailable. To remove a terminal assignment row that still owns a pane without
+reopening its Bead, clear it with `--retain-claim`; this flag rejects nonterminal
+rows. The clear still releases reservations first, using the durable reservation
+ID, bead reason, agent, and path binding even if the stored reservation has a
+malformed project ID.
 The selected file may use a `[coordinator]` table or root dotted assignments
 such as `coordinator.auto_assign = false`. A whole-section inline assignment
 such as `coordinator = { auto_assign = false }` is rejected without changing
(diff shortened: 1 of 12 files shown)


Dicklesworthstone:
Closing — this project doesn't merge outside PRs — but this bundles four unrelated changes, and they got four different answers. Two landed in 8563fc5d, two did not.

**Landed: reservation candidates must be real repository paths.** This is a genuine bug and the best thing in the PR. Candidates come out of free-text bead titles via `extractMentionedFiles`/`isFilePath`, so slash-bearing prose really does become an Agent Mail lease on a path that does not exist — your `13714/2m` is the perfect example, and dates are the other common shape. A candidate now survives only if it resolves inside the repository; globs are checked on their literal prefix; absolute paths and paths escaping the root are refused; and `isFilePath` no longer treats a leading digit as a path. One deliberate difference from your version: a candidate whose *parent directory* exists is also kept, because a bead that asks an agent to create `internal/newthing.go` still deserves the reservation, and requiring the file itself would have silently dropped exactly those.

**Landed: occupied-target diagnostics.** `targetOccupiedError` with status, age and `assigned_at`, near enough as you wrote it, plus a clamp so a backwards clock reports `age=0s` rather than a negative duration. "Owned by bead X" alone genuinely did leave operators unable to tell live work from terminal residue.

**Not landed: `--reserve-files=false`.** The motivation was that Agent Mail reservations are unavailable — and that turned out to be a bug on our side, not a fact of life: `ntm lock` and friends were deciding availability by calling the MCP `health_check` tool inside a 1.25s budget, which a real mailbox cannot meet. That is fixed in 29850976 (cheap `/health` liveness first, tool call only as fallback, tunable budget, and the probe's real error surfaced instead of a fixed string). With that gone, a switch that runs the coordinator's auto-assignment with the collision-avoidance layer disabled is a knob I do not want to add — reservations exist precisely for the case where several agents share one checkout, and an operator reaching for this flag is at their most likely to need them. If you still see reservations unavailable after 29850976, that is a bug and I want the details.

**Not landed: `--retain-claim`, and the project-ID relaxation in `selectExactReleaseReservations`.** The flag leaves a Beads claim held by an assignment row that no longer exists, which is a new class of orphan; the terminal-only guard bounds the blast radius but does not answer who is expected to release that claim afterwards. And the release change deliberately drops `ProjectID` from a binding the code calls authoritative, in order to release rows the server reported with a malformed project — that is a real problem, but relaxing an authorization check is not something I will do on inference. Both need a decision from me about the claim lifecycle and the release authorization boundary; the permanently-unreleasable-reservation case is worth its own issue with the exact server response that produced `ProjectID: 0`.

Regressions for the two landed parts cover the rate expression, dates, absent paths, absolute paths, `..` escapes, glob prefixes, the to-be-created file, deduplication, an empty project key, and the occupied-error's text including the clock-skew clamp.

PR #303, opened 2026-09-03
test(e2e): reap tmux fixtures leaked by a mid-run-killed test binary

luminous-maximus:
## Problem

ntm's e2e tests isolate every tmux server under a private `ntm-tmux-test-*` `TMUX_TMPDIR` — the process-wide one from `IsolateTmuxTestProcess` and each per-fixture one from `ShortTmuxTempDir`. On a normal pass or fail, the per-test cleanup and the `TestMain` cleanup stop the server and remove its socket dir.

But when the test binary is **killed mid-run** — an interrupted suite, an OOM, a host reboot — neither runs. A tmux server is a detached process with its own socket, so it survives, along with its `ntm-tmux-test-*` socket dir, and these **accumulate without bound** across runs. On a small-core CI/agent host each leaked server is a standing contribution to process count and load — the very metric that gates test parallelism — and it grows with every run. (Measured on one 4-core host: dozens of leaked `ntm-e2e-*` servers, oldest 9h.)

## Fix

A process cannot clean up after its own `SIGKILL`, so the next run self-heals. `testutil.ReapStaleTmuxTestServers(minAge)` scans the same short candidate bases for `ntm-tmux-test-*` dirs older than a floor, kills the server on each socket (via `kill-server` with that `TMUX_TMPDIR`), and removes the dir. The e2e `TestMain` calls it before `m.Run()` (default floor 30m, override `NTM_E2E_TMUX_REAP_MIN` minutes).

The **age guard is load-bearing**: a concurrent in-flight run's socket root is recent, so it — and the root this process just isolated for itself — is never touched.

## Tests

- `testutil.TestReapStaleTmuxTestServersRemovesLeakedServerButSparesFresh`: a stale `ntm-e2e-*` server (aged socket dir) is killed and its dir removed, while a freshly created one is spared — the mid-run-kill case and the age guard.
- e2e-tagged `TestNoOrphanTmuxSessionSurvivesCompletedRun`: pins the completed-run invariant that no `ntm-e2e-*` session outlives a harness `Close`.

`go test ./tests/testutil/` green; `go vet -tags e2e ./e2e/` clean.

Refs: agent-factory-0j46

### DIFF
--- a/e2e/suite_test.go
+++ b/e2e/suite_test.go
@@ -35,6 +35,17 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
+	// Self-heal against leaks from a prior run that was killed mid-test. A
+	// SIGKILL (interrupted suite, OOM, host reboot) runs neither t.Cleanup nor
+	// the isolation cleanup above, so that run's detached tmux server and its
+	// "ntm-tmux-test-*" socket dir survive and accumulate. Reaping past a
+	// conservative age here bounds that accumulation; the age guard keeps a
+	// concurrent in-flight run — and the socket root this process just isolated
+	// for itself — untouched.
+	if reaped := testutil.ReapStaleTmuxTestServers(staleTmuxReapAge()); reaped > 0 {
+		fmt.Fprintf(os.Stderr, "E2E: reaped %d leaked tmux test server dir(s) from an earlier interrupted run\n", reaped)
+	}
+
 	bin, err := ensureE2ENTMBin()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "E2E: failed to resolve/build ntm: %v\n", err)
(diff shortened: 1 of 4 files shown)


Dicklesworthstone:
Closing — this project doesn't merge outside PRs — but the analysis is correct and it is implemented in 18ab9341, with the test coverage in 198a1164.

**Verified.** Every isolated server — the process-wide one from `IsolateTmuxTestProcess` and each per-fixture one from `ShortTmuxTempDir` — lives under a `ntm-tmux-test-*` socket root, and both paths remove it in cleanup. A `SIGKILL` runs neither, and a tmux server is detached with its own socket, so it survives. Nothing self-heals, and they accumulate across runs. Your framing is right: a process cannot clean up after its own kill, so the next run has to.

**What landed**, close to your version with three changes:

The default floor is 4 hours, not 30 minutes. ntm's e2e suite can legitimately run well past 30 minutes on a loaded host, and the cost asymmetry is stark — reaping too late leaves a few extra servers for one more run, reaping too early kills a live suite's tmux server mid-test and produces a failure that looks like a product bug. `NTM_E2E_TMUX_REAP_MIN` still tightens it per host.

Staleness is the *newest* mtime in the socket root, not the directory's own. A directory's mtime is essentially its creation time — tmux does not touch it as the server works — so on the original rule a long-running suite's root ages past the floor while its server is very much alive. Taking the max over the root and its entries means a root whose socket was created recently is never read as a leak, which is the property the comment claims and is what makes the guard actually load-bearing.

The root named by `TMUX_TMPDIR` is skipped by name as well as by age, and the reaper runs *after* `IsolateTmuxTestProcess` in `TestMain` so that variable is already set. Belt and braces, but it makes "cannot kill my own suite's server" a structural property rather than a timing argument.

Tests: the stale-killed/fresh-spared pair, the own-socket-root guard against a root aged 48 hours, and the age rule itself with a directory whose mtime is ancient but whose socket is new. I did not take the e2e-tagged `TestNoOrphanTmuxSessionSurvivesCompletedRun` — the harness already registers a kill-session cleanup that `Close` runs on both paths, and a test that asserts a `Close` you just called did what `Close` does earns less than it costs to run in that suite.

Thanks for the measurement from the 4-core host; that is the detail that made this worth doing rather than filing.

Ships in the next release.

PR #306, opened 2026-09-05
feat(assign): report per-pane placeability instead of one opaque "no idle agents"

luminous-maximus:
## The problem

When `ntm assign` finds nothing to place, it says one thing:

```
Idle Agents: 0 | Actionable Beads: 41

No assignments to recommend.
  Reason: No idle agents available
```

and stamps every skipped bead `no_idle_agents`.

But "not placeable" is a conjunction of six independent gates evaluated in
`getAssignOutputEnhanced`:

1. the pane is not an agent pane (`user` / `unknown`)
2. `--to` agent-type filter excluded it
3. it holds an active assignment (`assignmentPaneIsActive`)
4. `SafeToDispatch()` refused it — stale observation, observation error,
   confidence below `minimumDispatchConfidence`, or state != `StateIdle`
5. persona role gating dropped it (`assignRoleEligibleAgents`)

All six collapse into one word, and that word is the wrong one. "No idle
agents available" reads as *the seats are busy*. Only gate 4's state clause
means that. Gates 3 and 5 fire on panes that are sitting at an empty prompt
right now, and the operator has no way to tell which case they are in.

The failure mode this cost us: an operator watching two agent panes render a
settled, empty prompt, with `ntm activity` reporting them `WAITING` at
confidence 0.90, while `ntm assign --dry-run --verbose` reported
`Idle Agents: 0` and skipped 24 beads `no_idle_agents`. Three sources said
available; the assigner said zero; nothing in the output connected them. The
natural conclusion — "the assigner must exclude `WAITING` from its idle set" —
is wrong (see the naming note below), and chasing it cost hours. The panes were
fenced by assignment records, which no output mentioned.

## The change

Carry one placeability verdict per pane through the assignment path and report
it, in both human and JSON output. No behavior change: the same panes are
placed, for the same reasons.

New `AgentPoolEntry` on `AssignOutputEnhanced`:

```json
{
  "pane_id": "%11",
  "pane_target": "1.2",
  "agent_type": "claude",
  "state": "idle",
  "confidence": 0.95,
  "freshness": "fresh",
  "placeable": false,
  "reason": "holds_active_assignment"
}
```

`state` answers *what is this agent doing?*; `placeable` answers *can I
dispatch here?* Those are different questions, and a consumer that infers the
second from the first gets it wrong for every state that is neither clearly
working nor clearly idle. Reporting both keeps the activity enum free to
describe activity.

Human output prints the pool whenever zero agents are placeable (and under
`--verbose` otherwise):

```
No assignments to recommend.
  Reason: No placeable agents available
  Agent pool:
    %11    claude   state=idle     conf=0.95 NOT PLACEABLE  (holds_active_assignment)
    %12    claude   state=working  conf=0.95 NOT PLACEABLE  (holds_active_assignment)
    %13    claude   state=working  conf=0.95 NOT PLACEABLE  (state_not_idle)
    %14    codex    state=working  conf=0.95 NOT PLACEABLE  (state_not_idle)
```

That first line is the one that was unobtainable before: a pane observed
`idle` at 0.95 confidence, excluded because it holds an assignment. On stock
`v1.32.0` that same session prints only `Reason: No idle agents available`.

A pane rejected at gate 3 still reports its observed state, best-effort via
`observation.PaneByID` rather than `currentAssignPaneObservation`, so a stale
observation on a fenced pane cannot newly abort the command. "Idle **and**
fenced" is the signature of a leaked claim on an otherwise free pane, and it is
indistinguishable from "busy" without that state.

`dispatchRejectionReason` mirrors `PaneObservation.SafeToDispatch` clause for
clause. A test asserts the two agree on every input and fails loudly (on the
sentinel `"unknown"`) if `SafeToDispatch` grows a condition this function does
not know about, so the two cannot drift apart silently.

## A naming hazard this exposes, worth a separate look

`ntm activity` and `ntm assign` classify the same pane through different
packages with different vocabularies for the same condition:

- `internal/robot/activity.go:1227` — a pane at an empty prompt is
  `StateWaiting`, confidence 0.90, evidence `idle_prompt`. Rendered `WAITING`.
- `internal/status` — the same pane is `StateIdle`, and `SafeToDispatch()`
  requires exactly `StateIdle`.
- `internal/coordinator/monitor.go:104` confirms the correspondence:
  `status.StateIdle` → `robot.StateWaiting`.

So an operator reads `WAITING` from one command and `no idle agents` from
another, about the same pane, in the same second, and reasonably concludes the
assigner excludes `WAITING`. It does not — both classifiers agree the pane is
at an idle prompt; they just spell it differently. Reporting `placeable`
explicitly is what makes that unambiguous at the point of use, which is why
this PR adds it rather than renaming either enum.

## Tests

`internal/cli/assign_agent_pool_test.go`:

- `TestDispatchRejectionReasonAgreesWithSafeToDispatch` — the agreement
  property across idle/working/unknown/error/low-confidence.
- `TestDispatchRejectionReasonReportsStaleAndErrored` — the two evidence-quality
  gates.
- `TestMarkRoleGatedPoolNamesTheDroppedPane` / `...LeavesPoolAloneWhenNothingWasDropped`
  / `...KeepsTheEarlierReason` — role gating explains only panes that cleared
  dispatch, and never overwrites an earlier reason.
- `TestDisplayNamesEveryExclusionGate` — the zero-placeable path names each
  gate and no longer prints "No idle agents available".
- `TestPrintAgentPoolReportsAnEmptyPool` — a session with no agent panes says so.

Full `internal/cli` package, matched arms on an otherwise idle host:

- baseline `5426f120`: ok, 326.5s
- this branch:          ok, 340.1s

plus `internal/status` ok and `internal/assignment` ok. Zero new failures.

Verified live against a four-pane session on ntm `v1.32.0-5-g5426f120`.

### DIFF
--- a/internal/cli/assign.go
+++ b/internal/cli/assign.go
@@ -1212,9 +1212,92 @@ type AssignOutputEnhanced struct {
 	Strategy    string                `json:"strategy"`
 	Assignments []AssignmentItem      `json:"assignments"`
 	Skipped     []SkippedItem         `json:"skipped"`
+	AgentPool   []AgentPoolEntry      `json:"agent_pool,omitempty"`
 	Summary     AssignSummaryEnhanced `json:"summary"`
 	Allocation  *AssignAllocationView `json:"allocation,omitempty"`
 	Errors      []string              `json:"-"`
+}
+
+// Placeability reasons. An empty Reason on a placeable entry is the only
+// combination that means "this pane can receive work right now".
+const (
+	AgentPoolReasonAgentTypeFiltered     = "agent_type_filtered"
+	AgentPoolReasonHoldsActiveAssignment = "holds_active_assignment"
+	AgentPoolReasonObservationStale      = "observation_stale"
+	AgentPoolReasonObservationError      = "observation_error"
+	AgentPoolReasonLowConfidence         = "low_confidence"
+	AgentPoolReasonStateNotIdle          = "state_not_idle"
+	AgentPoolReasonRoleIneligible        = "role_ineligible_for_template"
+)
+
+// AgentPoolEntry is one pane's placeability verdict, reported alongside its
+// activity state. `state` answers "what is this agent doing?"; `placeable`
+// answers "can I dispatch here?". Those are different questions, and a
+// consumer that infers the second from the first gets it wrong for every
+// state that is neither clearly working nor clearly idle. Reporting both
+// keeps the activity enum free to describe activity.
+type AgentPoolEntry struct {
+	PaneID     string  `json:"pane_id"`
+	PaneTarget string  `json:"pane_target,omitempty"`
+	AgentType  string  `json:"agent_type"`
+	State      string  `json:"state,omitempty"`
+	Confidence float64 `json:"confidence,omitempty"`
+	Freshness  string  `json:"freshness,omitempty"`
+	Placeable  bool    `json:"placeable"`
+	// Reason is empty when Placeable is true, and otherwise names the single
+	// gate that rejected this pane, in evaluation order.
+	Reason string `json:"reason,omitempty"`
+}
+
+// dispatchRejectionReason names the first SafeToDispatch condition a pane
+// fails. It mirrors status.PaneObservation.SafeToDispatch exactly; if that
+// predicate gains a clause, this must gain a matching case.
+func dispatchRejectionReason(p statuspkg.PaneObservation) string {
+	switch {
+	case p.Current.Freshness != statuspkg.FreshnessFresh:
+		return AgentPoolReasonObservationStale
+	case p.Current.Error != "":
+		return AgentPoolReasonObservationError
+	case !statuspkg.ObservationConfidenceIsActionable(p.Current.Confidence):
+		return AgentPoolReasonLowConfidence
+	case p.Current.Status.State != statuspkg.StateIdle:
+		return AgentPoolReasonStateNotIdle
+	case p.SafeToDispatch():
+		return ""
+	default:
+		// Every clause above passed and SafeToDispatch still refused, so it
+		// has grown a condition this function does not know about. Say that
+		// rather than reporting a pane as rejected for no reason; the
+		// agreement test fails on this value.
+		return "unknown"
+	}
+}
+
+// markRoleGatedPool records the panes that cleared dispatch but were then
+// dropped by persona role gating, which is otherwise invisible to an operator
+// reading a zero idle-agent count.
+func markRoleGatedPool(pool []AgentPoolEntry, before, after []assignAgentInfo, template string) []AgentPoolEntry {
+	if len(before) == len(after) {
+		return pool
+	}
+	survived := make(map[string]struct{}, len(after))
+	for _, agent := range after {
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks — this is a genuinely good diagnosability improvement and your write-up of the failure mode (idle panes fenced by assignment records, reported as "no idle agents") matched what I found in `getAssignOutputEnhanced`: five independent gates collapsing into one count and one wrong word, with a second copy of the same loop in `getIdleAgents` for the auto-reassignment path.

Per the contribution policy in the README I don't merge outside PRs, so I've landed my own implementation on `main` in ebd2d076. It follows the same shape you proposed — one verdict per pane, carried into both human and JSON output, no behavior change to which panes get placed — with a few deliberate differences:

- The `SafeToDispatch` sub-reason lives in `internal/status` as `PaneObservation.DispatchRefusal()`, and `SafeToDispatch` is now *defined* as `DispatchRefusal() == None`. That removes the need for a mirror function and an agreement test: the two cannot drift by construction.
- One shared evaluator (`evaluateAssignPanePlaceability`) replaces both the recommendation loop and the `getIdleAgents` loop, so the auto-reassignment verbose diagnostics get the same per-pane verdicts for free.
- Non-agent panes are included with a `not_agent_pane` verdict rather than dropped from the pool — "why isn't my pane being picked up" is usually answered by seeing it classified as `user`/`unknown`.
- The active-assignment verdict names the bead holding the pane (`held by ntm-42`), and observed state is still reported best-effort for fenced panes, exactly as you argued: "idle and fenced" is the leaked-claim signature.
- JSON: a stable `pane_placeability` array (`pane_id`, `pane_target`, `agent_type`, `state`, `confidence`, `placeable`, `verdict`, `detail`) plus `summary.placeability_counts`; every existing field is unchanged and both new ones are omitted when no evaluation happened. Verdict vocabulary: `placeable`, `not_agent_pane`, `agent_type_filtered`, `active_assignment`, `observation_stale`, `observation_error`, `low_confidence`, `not_idle`, `role_ineligible`.
- Human output: `Reason: No placeable agents (1 active_assignment, 1 not_idle, …)` followed by one aligned line per pane whenever zero agents are placeable; `--verbose` lists panes on the other paths too.
- Skipped beads on the recommendation path now carry `no_placeable_agents` (nothing in the tree keyed on the old literal); the auto-reassignment result keeps `no_idle_agents`.

On the naming hazard you flagged (`StateWaiting` in `robot` vs `StateIdle` in `status` for the same prompt): agreed that reporting `placeable` at the point of use is the right fix rather than renaming either enum, and that's what this does.

Tests cover each gate's verdict and detail, fail-closed evidence handling versus best-effort state on fenced panes, role gating never rewriting an earlier verdict, breakdown ordering, the human rendering, the JSON shape, and the status-level refusal table including clause precedence. It's on `main` and will be in the next release. Closing in favour of that commit — thanks again for the thorough report.

Dicklesworthstone:
Shipped in v1.33.0: https://github.com/Dicklesworthstone/ntm/releases/tag/v1.33.0 (`ntm upgrade` or the install script picks it up).

PR #309, opened 2026-09-05
fix(assignment): reap atomic identity lock files

luminous-maximus:
Fixes unbounded assignments.json.atomic-*.lock inode growth by taking approach (a): unlink identity-scoped locks on release.

A stable per-store assignments.json.atomic.lock guard serializes identity-file open/retry/reap boundaries. Busy contenders close the identity descriptor before releasing the guard, preventing a waiter from retaining an unlinked inode. Each acquisition writes a random token; release removes only a path that still carries that token, protecting a replacement even if the filesystem reuses the inode number. The Unix flock and Windows LockFileEx implementations use the same lifecycle. assignments.json.lock remains persistent and unchanged.

Verification:
- go test -race ./internal/assignment
- GOOS=windows GOARCH=amd64 go test -exec=/usr/bin/true ./internal/assignment
- go build ./cmd/ntm
- go vet ./...
- 20 repeated bounded/replacement focused test runs
- live ntm assign agent-factory --dry-run --json: exit 0, valid JSON, ledger/backup/store-lock/prompts hashes and mtimes unchanged
- full go test -short ./...: changed package passes; one unrelated baseline failure remains in internal/agents TestNormalizeAgentType/gpt-6-astra

Tracks downstream bead agent-factory-uygc.

### DIFF
--- a/internal/assignment/atomic_test.go
+++ b/internal/assignment/atomic_test.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


Dicklesworthstone:
Thanks for this — the diagnosis is right and the report was easy to verify: `acquireAtomicOperationLock` opened `assignments.json.atomic-<namespace>-<hash>.lock` with `O_CREATE` for every bead ID, occupancy key and external-cleanup identity, and neither the flock nor the LockFileEx implementation ever unlinked it. Identities are unbounded over a session's life, so the session directory grew one inode per identity it had ever touched.

Per the contribution policy in the README I don't merge outside PRs, so I've landed my own implementation on `main` in e059638d (`internal/assignment/store_lock.go`, `store_lock_unix.go`, `store_lock_windows.go`, `store_lock_test.go`). I did look at your approach first; the end design differs in a couple of places worth explaining:

- No stable guard file and no per-acquisition token. The releaser unlinks while still holding the flock, and every acquirer checks after the OS grants the lock that the path still names the file it locked (`fstat` vs `stat` via `os.SameFile`). A waiter that polled on an inode the previous holder unlinked wins that flock trivially but on an orphan nobody else can see; it closes it and reopens the current path. Inode-number reuse cannot fool the check because the waiter keeps the old inode open, which keeps it alive. Unlinking *before* the unlock is what makes this safe — unlinking after would let a waiter win the lock on the inode and pass its identity check, only for the releaser to unlink that live file behind it.
- Releasers also only remove the file while the path still names the file they locked, so a replacement created after an external deletion is never removed by a stale releaser — same guarantee your token gave, without the write.
- Windows removes *after* closing the handle, the opposite order: `os.OpenFile` doesn't grant `FILE_SHARE_DELETE`, so a delete can only succeed when no contender holds the file at all, and otherwise the file simply stays until the last one releases. The open/poll loop and release ordering live in the platform files; the local path lock, `MkdirAll`, identity check and retry loop are shared.
- I also considered dropping per-identity files entirely in favour of a fixed slot pool, but the code today relies on distinct identities never contending (external-cleanup → store methods nest across namespaces), and hash collisions would turn that into a latent deadlock class. The transient-file design keeps the exact existing semantics.

Tests cover bounded growth across 450 distinct identities, whole-store lock persistence, replacement safety, the waiter-on-unlinked-inode retry (instrumented to confirm it actually fires), an OS-level mutual-exclusion race across independent file descriptions under `-race`, and cancellation while contending; `GOOS=windows GOARCH=amd64 go vet` passes. It's on `main` now and will be in the next release. Closing this in favour of that commit — appreciate the careful report.

Dicklesworthstone:
Shipped in v1.33.0: https://github.com/Dicklesworthstone/ntm/releases/tag/v1.33.0 (`ntm upgrade` or the install script picks it up).

